PR #2808, opened 2026-04-26
feat: auto-compact and retry on context window errors

TheArchitectit:
## Summary

Automatically compact the session and retry when the model API returns a context window error. Users no longer need to manually run `/compact` when hitting context limits — recovery happens transparently with progressive compaction.

### Context window errors with no recovery path
Before: when a conversation grew too large for the model's context window, the API would return an error (e.g. `context_window_blocked`) and the turn would fail. The user had to manually run `/compact` and retry — if the session was still too large after one compaction, they were stuck. There was no automatic recovery and no gradual escalation.

**Fix:** The progressive auto-compact retry loop automatically detects context window errors and compacts the session in up to 4 rounds with decreasing `preserve_recent_messages` (4→2→1→0), trading conversation continuity for a smaller payload until the request fits. Each round:
1. Compacts the session via the Trident pipeline + summary compaction
2. Saves the compacted session back to `self.runtime` (critical — without this, `prepare_turn_runtime` would rebuild from the original un-compacted session)
3. Retries the request
4. If it still fails with a context window error, loops to the next round with more aggressive compaction

### Compacted session never actually used (Bug fix)
Before: the original auto-compact code called `prepare_turn_runtime(true)` after compaction, but `prepare_turn_runtime` reads from `self.runtime.session()` — the **original** un-compacted session. The compacted result was computed and discarded, so every retry sent the same bloated request. Compaction was completely wasted.

**Fix:** `*self.runtime.session_mut() = result.compacted_session.clone()` is called before rebuilding the runtime, so the retry uses the actually-compacted session.

### Panic when `preserve_recent_messages=0` (Bug fix)
Before: calling `compact_session` with `preserve_recent_messages=0` caused `saturating_sub(0) = messages.len()`, which produced `keep_from = messages.len()` — an out-of-bounds index when the code later accessed `session.messages[k]`.

**Fix:** explicit guard returns `messages.len()` for `preserve_recent_messages == 0`, which tells compaction "remove everything" without hitting the out-of-bounds path.

### Context overflow disguised as 400
Before: some OpenAI-compat backends return `400 "no parseable body"` when the request is too large to parse, instead of a proper `context_length_exceeded` error. The retry loop only checked for `"context_window"` in the error message, so these disguised context overflows were never caught.

**Fix:** the error detection now also matches `"no parseable body"`, so the progressive compact-retry loop triggers correctly for these cases too.

### Trident 3-stage compaction pipeline (new)

Before compaction was a single summary-based pass. Trident adds three structural reduction stages **before** summary compaction, making compaction much more aggressive:

| Stage | Name | What it does | Why |
|-------|------|--------------|-----|
| 1 | **Supersede** | Removes obsolete file operations (old reads/edits superseded by later ones on the same file) | Reading a file 5 times and editing it 3 times leaves 8 messages; only the latest read and edit matter |
| 2 | **Collapse** | Merges consecutive chatty exchanges (back-and-forth tool calls) into single summary messages | A 10-message debugging cycle (read→edit→read→edit…) becomes 1–2 messages |
| 3 | **Cluster** | Groups semantically similar messages using Jaccard fingerprinting | Multiple "fix lint errors" turns get merged into one summary |

After Trident runs, the existing summary-based compaction runs on the already-reduced session for maximum context reduction.

### Progressive auto-compact retry (new)

When a context window error is detected:
```
Round 1: preserve 4 recent messages → compact → retry
  ├─ Success → done
  └─ Still too large →
Round 2: preserve 2 recent messages → compact → retry
  ├─ Success → done
  └─ Still too large →
Round 3: preserve 1 recent message → compact → retry
  ├─ Success → done
  └─ Still too large →
Round 4: preserve 0 recent messages (maximum compaction) → compact → retry
  ├─ Success → done
  └─ Still too large → surface error to user
```

## Files changed

| File | What changed |
|------|-------------|
| `runtime/src/trident.rs` | **New file** (~791 lines). 3-stage Trident pipeline (supersede/collapse/cluster), config, stats, 6 tests |
| `runtime/src/compact.rs` | Fix panic when `preserve_recent_messages=0` — guard against out-of-bounds index |
| `runtime/src/lib.rs` | Export `trident` module |
| `rusty-claude-cli/src/main.rs` | Progressive auto-compact retry loop with `session_mut` fix, `/compact` updated to use Trident, `"no parseable body"` detection |

## Test plan

- [x] `cargo test --workspace` — all tests pass
- [x] 6 Trident unit tests pass (supersede, collapse, cluster, full pipeline, stats)
- [x] Tested end-to-end with bloated session against OpenAI-compat provider — progressive auto-compact retry triggered and completed successfully
- [ ] Run a session that exceeds the model's context window — automatic compaction + retry occurs
- [ ] Non-context-window errors still propagate without triggering auto-compact
- [ ] Session is persisted to disk after successful auto-compact retry

💘 Generated with Crush

### DIFF
--- a/rust/crates/api/src/providers/openai_compat.rs
+++ b/rust/crates/api/src/providers/openai_compat.rs
@@ -790,6 +790,7 @@ struct ChatCompletionChunk {
 
 #[derive(Debug, Deserialize)]
 struct ChunkChoice {
+    #[serde(default)]
     delta: ChunkDelta,
     #[serde(default)]
     finish_reason: Option<String>,
--- a/rust/crates/api/src/providers/openai_compat.rs
+++ b/rust/crates/api/src/providers/openai_compat.rs
@@ -1353,7 +1353,50 @@ fn parse_sse_frame(
             data_lines.push(data.trim_start());
         }
     }
+    // If no SSE data lines found, check if the entire frame is raw JSON (error or otherwise)
     if data_lines.is_empty() {
+        // Detect raw JSON error response (not SSE-framed)
+        if let Ok(raw) = serde_json::from_str::<serde_json::Value>(trimmed) {
+            if let Some(err_obj) = raw.get("error") {
+                let msg = err_obj
+                    .get("message")
+                    .and_then(|m| m.as_str())
+                    .unwrap_or("provider returned an error")
+                    .to_string();
+                let code = err_obj
+                    .get("code")
+                    .and_then(serde_json::Value::as_u64)
+                    .map(|c| c as u16);
+                let status = reqwest::StatusCode::from_u16(code.unwrap_or(500))
+                    .unwrap_or(reqwest::StatusCode::INTERNAL_SERVER_ERROR);
+                return Err(ApiError::Api {
+                    status,
+                    error_type: err_obj
+                        .get("type")
+                        .and_then(|t| t.as_str())
+                        .map(str::to_owned),
+                    message: Some(msg),
+                    request_id: None,
+                    body: trimmed.chars().take(500).collect(),
+                    retryable: false,
+                    suggested_action: suggested_action_for_status(status),
+                    retry_after: None,
+                });
+            }
+        }
+        // Detect HTML responses
+        if trimmed.starts_with('<') || trimmed.starts_with("<!") {
+            return Err(ApiError::Api {
+                status: reqwest::StatusCode::BAD_REQUEST,
+                error_type: Some("invalid_response".to_string()),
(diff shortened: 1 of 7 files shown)


Pilser:
Yeah we need this , and if possible some one to add fff search [ ([dependencies]
fff-search = "0.6") ](https://github.com/dmtrKovalenko/fff.nvim) ,  the Issue page is to messy to propose feature request

TheArchitectit:
This fixes the 400 error if you run out of compact space.

PR #2810, opened 2026-04-26
feat: interactive provider wizard (/setup, claw setup, Ctrl+P)

TheArchitectit:
## Summary

Adds an interactive provider wizard so users can configure their AI provider, API key, base URL, and model without manually editing config files or setting environment variables. Settings persist across sessions in `~/.claw/settings.json`.

**Three ways to access the wizard:**
- `claw setup` — CLI subcommand
- `/setup` — slash command inside the REPL
- **Ctrl+P** — hotkey inside the REPL (inserts `[Provider Swap]` prompt, press Enter to confirm)

### No way to configure providers without environment variables
Before: the only way to set your provider, API key, base URL, or model was via environment variables (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `OPENAI_BASE_URL`, etc.) or manually editing a settings file. There was no guided setup — users had to read docs, figure out which env vars to set, and hope they got the names right. Switching providers mid-session required exiting, setting new env vars, and restarting.

**Fix:** The interactive wizard walks users through provider selection, API key entry, base URL (for OpenAI-compat), and model name — all with prompts and validation. Settings persist in `~/.claw/settings.json`, so they survive across sessions. Ctrl+P enables in-session provider hot-swap without restarting.

### No persistent auth — re-enter credentials every session
Before: API keys entered as environment variables are lost when the terminal closes. There was no way to save them persistently without editing shell profiles. Users who don't use `.env` files or shell rc files had to export keys every time.

**Fix:** The wizard saves all settings (including API key) to `~/.claw/settings.json` with `0o600` permissions. Auth resolution follows a three-tier priority: env var > `.env` file > stored config. Environment variables always win, so power users aren't blocked.

### Model written under wrong JSON key
Before: `save_user_provider_settings()` wrote `model` nested under the `provider` object in `settings.json`, but the config reader expected it at the top level. This meant the model selection from the wizard was silently ignored on next startup.

**Fix:** `model` is now written at the top level of `settings.json` alongside the `provider` block, matching what the config reader expects.

### Three-tier auth resolution

`read_env_or_config()` checks in order:
1. Environment variable (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, etc.)
2. `.env` file in the current directory
3. Stored config from `~/.claw/settings.json`

### How Ctrl+P works

Rustyline cannot chain commands or show popup menus. The implementation:
1. Ctrl+P inserts invisible sentinel character `\x01`
[...]

PR #2811, opened 2026-04-27
fix: /resume latest searches all workspaces

TheArchitectit:
## Summary

Fixes `/resume latest` not finding sessions. Two separate bugs that together made session resume nearly unusable when switching directories.

### Bug 1: `/resume latest` only searches current workspace
Before: sessions were only searched in the current workspace's fingerprinted directory (`<cwd>/.claw/sessions/<fingerprint>/`). If you started `claw` from a different directory than where your session was created, `/resume latest` found **zero** sessions even though sessions existed elsewhere on disk. Since sessions are partitioned by workspace fingerprint (hash of the CWD), switching directories made all previous sessions invisible.

This is especially painful for users who work in multiple project directories — switching from `~/project-a` to `~/project-b` means `/resume latest` in project-b can never find sessions from project-a.

**Fix:** Three-tier session lookup fallback in `latest_session()`:
1. **Tier 1**: Current workspace's fingerprint dir (existing behavior, unchanged)
2. **Tier 2**: Scan all fingerprint subdirs under the project-local `<cwd>/.claw/sessions/`
3. **Tier 3**: Scan all fingerprint subdirs under the global `~/.claw/sessions/`

`scan_global_sessions()` iterates all fingerprint subdirectories using `collect_sessions_from_dir_unvalidated()` which skips workspace validation. Results are sorted by recency.

### Bug 2: `/resume latest` returns the empty current session
Before: `/resume latest` always returned the most recently created session — which was the empty session just created on startup. So every time you ran `/resume latest` from a fresh `claw` instance, it would "resume" a brand new empty session instead of your previous conversation with actual history.

**Fix:** `load_session_loose()` skips:
- The current session ID (so `/resume latest` from a running REPL doesn't return itself)
- Sessions with `message_count == 0` (so empty startup sessions are skipped, returning the previous one with actual history)

The exclude ID flows from the REPL CLI through `load_session_reference()` → `SessionStore::load_session_loose()` → `resolve_reference()` → `latest_session()`.

### Cross-workspace resume

`load_session_loose()` allows cross-workspace resume for alias references (`latest`, `last`, `recent`), printing a note like:
```
  Note: resuming session from a different workspace (origin: /home/user/project-alpha)
```
Explicit session IDs still enforce workspace validation — you must be explicit about cross-workspace resume.

## Files changed

| File | What changed |
|------|-------------|
| `runtime/src/session_control.rs` | Added `global_sessions_root()`, `scan_global_sessions()` with 3-tier fallback, `load_session_loose()` for cross-workspace aliases, `collect_sessions_from_dir_unvalidated()`. Updated `latest_session()` to try global fallback. Updated error message hint. |
| `runtime/src/config_validate.rs` | Added `provider` top-level field + `PROVIDER_FIELDS` constant (shared with #2810 — needed because user's settings.json already contains the provider key) |
| `rusty-claude-cli/src/main.rs` | Use `load_session_loose()` for alias references so `/resume latest` can find sessions across workspaces |
| `commands/src/lib.rs` | `#[allow(clippy::unnecessary_wraps)]` on two helper functions that return `Result` for future-proofing |

## Test plan

- [x] `cargo test --workspace` — all tests pass
- [x] `/resume latest` in REPL — returns previous session with messages, not empty current session
- [x] Start `claw`, have a conversation, exit, start `claw` again, `/resume latest` — brings back the previous chat
- [x] Session message count and turn count shown correctly on resume
- [ ] `/resume <explicit-id>` from different workspace — still errors with WorkspaceMismatch
- [ ] `/resume latest` from a completely different directory — finds session across workspaces

💘 Generated with Crush

### DIFF
--- a/rust/crates/api/src/providers/openai_compat.rs
+++ b/rust/crates/api/src/providers/openai_compat.rs
@@ -790,6 +790,7 @@ struct ChatCompletionChunk {
 
 #[derive(Debug, Deserialize)]
 struct ChunkChoice {
+    #[serde(default)]
     delta: ChunkDelta,
     #[serde(default)]
     finish_reason: Option<String>,
--- a/rust/crates/api/src/providers/openai_compat.rs
+++ b/rust/crates/api/src/providers/openai_compat.rs
@@ -1353,7 +1353,50 @@ fn parse_sse_frame(
             data_lines.push(data.trim_start());
         }
     }
+    // If no SSE data lines found, check if the entire frame is raw JSON (error or otherwise)
     if data_lines.is_empty() {
+        // Detect raw JSON error response (not SSE-framed)
+        if let Ok(raw) = serde_json::from_str::<serde_json::Value>(trimmed) {
+            if let Some(err_obj) = raw.get("error") {
+                let msg = err_obj
+                    .get("message")
+                    .and_then(|m| m.as_str())
+                    .unwrap_or("provider returned an error")
+                    .to_string();
+                let code = err_obj
+                    .get("code")
+                    .and_then(serde_json::Value::as_u64)
+                    .map(|c| c as u16);
+                let status = reqwest::StatusCode::from_u16(code.unwrap_or(500))
+                    .unwrap_or(reqwest::StatusCode::INTERNAL_SERVER_ERROR);
+                return Err(ApiError::Api {
+                    status,
+                    error_type: err_obj
+                        .get("type")
+                        .and_then(|t| t.as_str())
+                        .map(str::to_owned),
+                    message: Some(msg),
+                    request_id: None,
+                    body: trimmed.chars().take(500).collect(),
+                    retryable: false,
+                    suggested_action: suggested_action_for_status(status),
+                    retry_after: None,
+                });
+            }
+        }
+        // Detect HTML responses
+        if trimmed.starts_with('<') || trimmed.starts_with("<!") {
+            return Err(ApiError::Api {
+                status: reqwest::StatusCode::BAD_REQUEST,
+                error_type: Some("invalid_response".to_string()),
+                message: Some("provider returned HTML instead of JSON (check endpoint URL)".to_string()),
+                request_id: None,
+                body: trimmed.chars().take(200).collect(),
+                retryable: false,
+                suggested_action: Some("verify the API endpoint URL is correct".to_string()),
+                retry_after: None,
+            });
+        }
         return Ok(None);
     }
     let payload = data_lines.join("\n");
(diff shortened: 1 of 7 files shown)


Pilser:
hello @TheArchitectit  , Can you also fix when xai  configured and ask who are you the  model respond that it is claude model yet it config it show xai is the one configured , and I dont have  claude key so its  xai model claim being  claude ,

I recently sent this,
 massage to @[programming-pupil](https://github.com/programming-pupil)
programming-pupil   this massage and he fixed only xai api to be compatible  but model identity still failed .
 
 here the old massage (Can you also fix this? ## xAI / Grok-3 Integration
Problem

claw always routed requests to api.anthropic.com regardless of which model was specified. Using --model grok-3 still sent requests to Anthropic, causing 401 errors.
Root Cause

AnthropicRuntimeClient.client was hardcoded to AnthropicClient, which always uses Authorization: x-api-key and sends to Anthropic's endpoint. ProviderClient already existed in the codebase and could route by model name prefix (grok-* → xAI, claude-* → Anthropic) but was never wired into the CLI runner.
Fix — claw-code-parity/rust/crates/rusty-claude-cli/src/main.rs

    Added ProviderClient to the use api::{...} import block
    Changed AnthropicRuntimeClient.client field type from AnthropicClient → ProviderClient
    Changed AnthropicRuntimeClient::new() to use ProviderClient::from_model_with_anthropic_auth(&model, Some(resolve_cli_auth_source()?))
    Updated push_prompt_cache_record signature from &AnthropicClient → &ProviderClient

Fix: claw Says "I'm Claude Opus 4.6" When Using Grok
Problem

When asked "what model are you?", claw replied "I'm Claude, specifically Claude 3.7 Sonnet" even when running grok-3.
Root Cause

runtime/src/prompt.rs:38:

pub const FRONTIER_MODEL_NAME: &str = "Claude Opus 4.6";

This was injected into every session's system prompt as:

Model family: Claude Opus 4.6

Grok reads its own system prompt and reports that as its identity. No Anthropic API key required — it's just hardcoded text.
Fix — claw-code-parity/rust/crates/runtime/src/prompt.rs

    Added model_name: Option<String> field to SystemPromptBuilder
    Added pub fn with_model(mut self, model_name: impl Into<String>) -> Self builder method
    Changed environment_section() to use self.model_name.as_deref().unwrap_or(FRONTIER_MODEL_NAME) instead of the constant
    Added pub fn load_system_prompt_for_model(...) — same as load_system_prompt but accepts model_name: Option<&str> and calls builder.with_model(name)

Fix — claw-code-parity/rust/crates/runtime/src/lib.rs

    Added load_system_prompt_for_model to the pub use prompt::{...} re-export block

Fix — claw-code-parity/rust/crates/rusty-claude-cli/src/main.rs

    Imported load_system_prompt_for_model from runtime
    Changed build_system_prompt() signature to build_system_prompt(model: &str)
    Passes Some(model) to load_system_prompt_for_model
    Updated LiveCli::new() to call build_system_prompt(&model)
)

TheArchitectit:
The other feature which allows you set to provider should take care of this?

ankurbakery (approved):

TheArchitectit:
what else do you need to merge this in?

1716775457damn:
Nice fix. The three-tier fallback (current workspace → project-local → global) is the right approach for this class of problem. Bug 2 with the empty current session being selected is particularly insidious — easy to miss in testing but frustrating in practice. Looks good.

PR #2814, opened 2026-04-27
fix: claw setup writes model at wrong JSON level

TheArchitectit:
## Summary

- Fix `claw setup` wizard writing `model` inside the `provider` object instead of at the top level of `settings.json`
- `parse_optional_model` reads from the top level, so the model setting was silently ignored after running `claw setup`
- Also remove the top-level `model` key when clearing provider settings

## Before

```json
{
  "provider": {
    "kind": "openai",
    "apiKey": "...",
    "baseUrl": "...",
    "model": "glm-5.1-fast"  ← ignored by parser
  }
}
```

## After

```json
{
  "model": "glm-5.1-fast",   ← read correctly by parser
  "provider": {
    "kind": "openai",
    "apiKey": "...",
    "baseUrl": "..."
  }
}
```

## Test plan

- [x] `cargo build --release` — clean build
- [ ] Run `claw setup`, select a provider and model, verify `model` appears at top level in `~/.claw/settings.json`
- [ ] Run `claw /status` and verify the model is reflected
- [ ] Run `clear_user_provider_settings` path and verify both `provider` and top-level `model` are removed

💘 Generated with Crush

### DIFF
--- a/rust/crates/api/src/providers/anthropic.rs
+++ b/rust/crates/api/src/providers/anthropic.rs
@@ -738,11 +738,7 @@ fn now_unix_timestamp() -> u64 {
 }
 
 fn read_env_non_empty(key: &str) -> Result<Option<String>, ApiError> {
-    match std::env::var(key) {
-        Ok(value) if !value.is_empty() => Ok(Some(value)),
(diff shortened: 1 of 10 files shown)


TheArchitectit:
Fix added to #2810 instead

Issue #336, opened 2026-03-31
ClaudeCode学习交流群


huchunnuan:
<img width="1031" height="1449" alt="Image" src="https://github.com/user-attachments/assets/8dfd3e0f-09d5-46da-b862-0cec0b1b3277" />

PR #2844, opened 2026-04-29
docs(roadmap): add #336 — session subcommand resume inconsistency and type/kind error mismatch

code-yeongyu:
## Pinpoint #329

**Gap 1:** `/session` vs `/session list` inconsistent resume-mode support.
**Gap 2:** Error JSON uses `"type"` instead of `"kind"` — breaks generic kind-dispatchers.

**Live on `b90875f`:**
```
claw --output-format json --resume latest /session
→ {"command":"/session","error":"unsupported resumed slash command","type":"error"}
                                                                         ^^^^^ not "kind"

claw --output-format json --resume latest /session list
→ valid session list JSON ✅
```

Two subcommands of same slash command, different resume contracts, no per-subcommand resume flag in `/help` JSON.

**Fix shape:** add `resume_subcommands` per-command in `/help` JSON; use `kind=error` in error responses; regression tests.

**ROADMAP.md only**


code-yeongyu:

1716775457damn:
Session subcommand resume inconsistency and type/kind error mismatch are worth tracking. The inconsistent behavior between resume paths makes debugging session state difficult. Documented on roadmap.

PR #2854, opened 2026-04-29
docs(roadmap): add #339 — session delete not resume-safe, blocks GC automation

code-yeongyu:
Pinpoint #339: /session delete not resume-safe while /session list is. ROADMAP.md only.


Yeachan-Heo:
Thanks for the report. I’m closing this roadmap-only PR rather than merging it as-is; the underlying session-GC gap is noted and can be handled through the implementation/backlog path with a concrete fix.

—

Issue #362, opened 2026-03-31
VibeCoding 爱好者来，点关注不迷路


CreateSun:
![Image](https://github.com/user-attachments/assets/4fc1ec27-e156-4428-ae63-19db1be8d35c)

PR #2865, opened 2026-04-29
docs(roadmap): add #362 — doctor auth false-positive: misses CLI session tokens

code-yeongyu:
## Pinpoint #345

`/doctor` auth check reports `warn` even when valid CLI session auth exists.

**Live on `f65b2b4`:**
```json
{"name":"auth","status":"warn","summary":"no supported auth env vars were found",
 "api_key_present":false,"auth_token_present":false}
```

Yet `claw --resume latest` works fine (valid session in ~/.claw/). Doctor only probes env vars + legacy refresh token, misses CLI session tokens and OAuth tokens in config files → false-positive warning on every healthy machine that uses browser/OAuth auth.

**ROADMAP.md only**


code-yeongyu:

1716775457damn:
Doctor auth false-positive is a concrete bug — if it misses CLI session tokens, users get misleading auth status reports. Should check all token sources consistently.

PR #2868, opened 2026-04-29
docs(roadmap): add #356 — session list title always null; no rename command

code-yeongyu:
## Pinpoint #347

`/session list` returns `"title": null` for all sessions — no human-readable label, no rename command, sessions distinguishable only by epoch-based ID.

**Live on `a1bfcd4`:**
```json
{"id": "session-1776891003038-0", "title": null, "message_count": 1, ...}
// all 10 sessions: title: null
```

Orchestration cannot build a session picker without reading every session's full message history. No `/session rename` command exists.

**ROADMAP.md only**


code-yeongyu:

1716775457damn:
Session list title always null is a real UX gap — users cannot identify sessions by name. A rename command would solve it. Documented on roadmap.

Issue #350, opened 2026-03-31
量化、A股相关Skill 加微信：endofworld42


Etherstrings:
need a test

PR #2889, opened 2026-04-29
docs(roadmap): add #350 — plugins enable nonexistent hangs 8s with 0 bytes output

code-yeongyu:
## Pinpoint #350

`/plugins enable <nonexistent> --output-format json` hangs 8+ seconds, produces 0 bytes.

**Live on `ee44ff98`:**
```
./claw --output-format json --resume latest /plugins enable does-not-exist
# timeout 8s: 0 bytes stdout, 0 bytes stderr, no error envelope
# repeat: same hang; third sample: stuck
```

Silent hang with no output. Breaks orchestration timeouts (typical 5-30s). No diagnostic info.

**ROADMAP.md only**


code-yeongyu:
Superseded.

1716775457damn:

PR #2893, opened 2026-04-29
docs(roadmap): add #365 — allowed_tools.entries null ambiguous; unrestricted vs unknown

code-yeongyu:
## Pinpoint #365

`/status` `allowed_tools.entries: null` is ambiguous — can't tell if it means all-tools-allowed or tool-list-unavailable.

**Live on `0f9e8915`:**
```json
{"allowed_tools":{"entries":null,"restricted":false,"source":"default"}}
```

With `entries: null`, automation can't distinguish: (a) unrestricted (all tools), (b) list not loaded, (c) enumeration failed. `restricted: false` helps but doesn't fully resolve the ambiguity.

**ROADMAP.md only**


code-yeongyu:
Superseded.

1716775457damn:

PR #2896, opened 2026-04-29
Add web-ui: tmux+xterm.js wrapper for the claw REPL

prcdslnc13:
## Summary

- Adds `web-ui/` — a localhost-only single-user web app that runs claw inside tmux per session. The browser renders an xterm.js terminal embedded in a structured shell (project + session pickers, PWA manifest framing).
- Multi-turn, approval prompts, and browser-disconnect resilience all happen at the terminal layer; tmux owns the long-lived REPL so disconnects don't kill in-flight work and server restarts don't lose claw state.
- A subprocess-per-turn fallback (regex-mediated approval, JSON output parsing, conversation-history injection) lives at `?ui=legacy` and will be deleted in a follow-up once the tmux path has more usage time.

## Why this shape

The first attempt was the subprocess path (parse `--output-format json`, regex-detect `Approve this tool call?` prompts, mediate via a UI modal, inject conversation history into each turn's prompt). It worked but every part was structurally fragile: claw's JSON schema isn't pinned, the prompt regex breaks on any TUI rendering change, and `--resume` is inspect-only so multi-turn required token-wasteful history injection.

Sentinel (`~/src/ai-local-hardware-deploy`) had already solved this for a similar use case: tmux owns the long-lived REPL, the browser uses xterm.js to render the PTY directly. claw runs natively, conversation state lives in claw's own memory, approval prompts are answered by typing `y` / `n` in the terminal. The pivot deletes the entire fragile mediation layer and replaces it with a thin PTY relay.

## Architecture

- `server/` — Python+FastAPI:
  - SQLite-backed project + session registry (`projects.py`, `sessions.py`, `db.py`).
  - `tmux_manager.py` — create / attach-pty / resize / kill tmux session per claw-web session.
  - `api.py` — REST CRUD + `/api/sessions/{id}/terminal` WebSocket relay (browser ↔ PTY).
  - `claw_runner.py` + legacy `/api/sessions/{id}/stream` WS — phase-1.5/1.6 fallback.
  - `mock_runner.py` — canned-response runner so the UI walks without claw plumbed.
- `static/` — Vanilla HTML/JS (no build step):
  - `xterm.js` + `xterm-addon-fit` from CDN.
  - Sidebar pickers, terminal panel, legacy structured panel, PWA manifest + service-worker stub.

## Test plan

- [ ] `cd web-ui && ./scripts/dev.sh` boots on port 7683 (mock mode default; `CLAW_WEB_MODE=subprocess` for real claw)
- [ ] Browser at http://127.0.0.1:7683 shows the project picker
- [ ] Create / pick a project, create a session → xterm.js terminal opens, claw boots inside tmux
- [ ] Multi-turn conversation threads context (claw stays alive between turns)
- [ ] Approval prompts answered with literal `y` / `n` keystrokes route through to claw
- [ ] Browser disconnect + reconnect picks up the same claw session, scrollback intact
- [ ] Mouse wheel scrolls tmux scrollback (mouse mode is enabled per session)
- [ ] \`?ui=legacy\` falls back to the subprocess-per-turn path with the approval modal
- [ ] \`Kill session\` button destroys the tmux session and starts a fresh REPL on reconnect
- [ ] \`pytest server/tests\` → 14 / 14 green


prcdslnc13:
Mistargeted PR — meant for fork prcdslnc13/claw-code, not upstream. Reopening on the correct repo. Sorry for the noise.

PR #2902, opened 2026-04-30
docs(roadmap): add #370 — resume non-slash-cmd error hint is null; constraint undocumented in help

code-yeongyu:
## Pinpoint #370

`--resume` only accepts slash commands as trailing args, but this isn't in `--help` and the error has `hint: null`.

**Live on `19947545`:**
```
$ claw --output-format json --resume latest session list
{"error":"--resume trailing arguments must be slash commands","hint":null,...}
```

`--help` shows `claw --resume latest /status /diff` but never says "only slash commands allowed". User tries `--resume latest doctor` instead of `--resume latest /doctor` — gets null hint with no actionable suggestion.

**Fix:** populate `hint` with corrected form when bare subcommand has slash-equivalent; document constraint in `--help`.

**ROADMAP.md only**


code-yeongyu:
Superseded.

1716775457damn:

Issue #371, opened 2026-03-31
Adding this to the OWASP Top 10 for LLM screwups


rajnandan1:
https://rajnandan.com/posts/owasp-top-10-ai-vulnerabilities/

PR #2903, opened 2026-04-30
docs(roadmap): add #371 — /cache returns kind=stats; /providers returns doctor response

code-yeongyu:
## Pinpoint #371

Two commands return wrong-kind/wrong-command responses entirely.

**Live on `19947545`:**
```json
// /cache:
{"kind":"stats", "cache_creation_input_tokens":0, ...}
// kind should be "cache", not "stats"

// /providers:
{"kind":"doctor", "checks":[...], "has_failures":false, ...}
// full doctor response — not a providers list at all
```

Both bounded-reproducible (≥2 samples). Distinct from #360 (identical blobs): here commands return wrong-kind or wrong-command responses entirely.

**ROADMAP.md only**


code-yeongyu:
Superseded.

1716775457damn:

PR #2946, opened 2026-04-30
docs(roadmap): add #404 — status allowed_tools entries:null ambiguous

code-yeongyu:
## Pinpoint #404

`status --output-format json` returns:
```json
"allowed_tools": {"entries": null, "restricted": false, "source": "default"}
```

`entries:null` is ambiguous — does it mean unrestricted (all tools), unset, or empty-blocking?
Per #97, `--allowedTools ""` produces empty BTreeSet blocking all tools. If that also emits `entries:null`, automation cannot detect a self-restricted lane from status JSON.

**ROADMAP.md only**


code-yeongyu:
Roadmap entries consolidated in direct push.

PR #2975, opened 2026-05-01
feat(auth): Add welcome screen and claw auth command for provider setup

bloodf:
## Summary

When opening claw without a configured provider API key, users now see an interactive welcome screen instead of a hard error. They can select a provider, enter their API key, and continue using claw immediately.

### Welcome Screen
- Automatically shown at REPL startup or prompt mode when auth is missing
- Lists built-in providers (Anthropic, OpenAI, xAI)
- Sets the env var for the current process so claw works immediately
- Optionally saves the model choice to ~/.claw/settings.json
- Supports cancellation (press Enter without entering a key)

### New CLI Command: `claw auth [PROVIDER]`
- Authenticate with a specific provider directly: `claw auth openai`
- Without a provider argument, shows the interactive picker
- Sets env vars for the current session

### Implementation
- Added `BuiltinProvider` struct and constants for built-in providers
- Added `check_model_auth_available()` to detect missing credentials
- Added `run_provider_welcome()` for the interactive onboarding flow
- Added `run_auth_command()` for the CLI subcommand
- Hooked welcome screen into `run_repl()` and `Prompt` mode dispatch
- Added `CliAction::Auth` variant and `parse_args` support
- Updated help text and typo-suggestion list
- Added unit tests for auth subcommand parsing

## Testing
- `cargo check --workspace` passes
- New unit tests: `parse_args_auth_without_provider`, `parse_args_auth_with_provider`

### DIFF
--- a/rust/crates/api/src/lib.rs
+++ b/rust/crates/api/src/lib.rs
@@ -18,9 +18,9 @@ pub use prompt_cache::{
     CacheBreakEvent, PromptCache, PromptCacheConfig, PromptCachePaths, PromptCacheRecord,
     PromptCacheStats,
 };
-pub use providers::anthropic::{AnthropicClient, AnthropicClient as ApiClient, AuthSource};
+pub use providers::anthropic::{has_auth_from_env_or_saved as anthropic_has_auth, AnthropicClient, AnthropicClient as ApiClient, AuthSource};
 pub use providers::openai_compat::{
-    build_chat_completion_request, flatten_tool_result_content, is_reasoning_model,
+    build_chat_completion_request, flatten_tool_result_content, has_api_key, is_reasoning_model,
     model_rejects_is_error_field, translate_message, OpenAiCompatClient, OpenAiCompatConfig,
 };
 pub use providers::{
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -24,10 +24,11 @@ use std::thread::{self, JoinHandle};
 use std::time::{Duration, Instant, UNIX_EPOCH};
 
 use api::{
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
REQUEST_CHANGES from dogfood/review on head `e46e1d33d6c221896ae23209ac0487ea6f189598`.

Blockers:

1. `cargo fmt --check` fails on the PR diff (`rust/crates/api/src/lib.rs`, `rust/crates/rusty-claude-cli/src/main.rs`).
2. `claw auth <provider>` reports successful authentication but only sets the API key in the current process with `std::env::set_var(...)` and then exits. A follow-up `claw --model openai/gpt-4o prompt hi` in the same temp HOME still has no credentials and fails auth again. That makes the new auth command look durable when it is not.
3. The API key prompt uses `read_line`, so the secret is echoed in the terminal in both `claw auth <provider>` and the welcome flow.

Dogfood evidence:
- Session: `claw-code-pr-2975-auth-welcome-review`
- `git diff --check origin/main...HEAD` passed.
- `cargo test -p rusty-claude-cli parse_args_auth -- --nocapture` passed (4 tests).
- `cargo test -p api resolve_startup_auth_source -- --nocapture` passed.
- `cargo test -p rusty-claude-cli local_subcommand_help_does_not_fall_through_to_runtime_or_provider_calls -- --nocapture` passed.
- `cargo fmt --check` failed.
- Temp-HOME dogfood: `claw auth openai` accepted `dogfood-test-key` and printed `Authentication set for OpenAI.`, but a new process with the same HOME still failed with `Authentication required...`.

Required delta: either persist auth intentionally and securely, or change the UX/copy so `claw auth` does not claim durable authentication; API-key input also needs hidden/no-echo handling.

—

bloodf:
Superseded by #2977 which combines provider profiles, auth, and OAuth into a single PR.

Issue #126, opened 2026-03-31
我打算成立Bnthropic


FishMooneyL:

1716775457damn:
The claw-code leak has certainly sparked creative reactions. This issue is closed - for serious discussions about open source code transparency, please use the project's discussion board.

PR #2990, opened 2026-05-04
fix(config): emit section and section_value in JSON output for config subcommands

code-yeongyu:
## Problem

`claw config model --output-format json` returned the same envelope as `claw config --output-format json` — the section argument was parsed but discarded (`_section` parameter in `render_config_json`).

```json
// before: config model → identical to config (no section field)
{ "kind": "config", "merged_keys": 1, "loaded_files": 1, "files": [...] }

// after: config model
{ "kind": "config", "section": "model", "section_value": null,
  "merged_keys": 1, "loaded_files": 1, "files": [...] }
```

The text path (`render_config_report`) already handled section correctly; only the JSON path was broken.

## Fix

`render_config_json` now:
- Passes `section` through (was `_section`)
- Looks up the section value via `runtime_config.get()`, converting the internal `JsonValue` to `serde_json::Value` via `render()`+parse
- Emits `section` (string) and `section_value` (parsed JSON value or null)
- Returns `ok:false` + `error` for unsupported section tokens

## Test

`config_section_json_emits_section_and_value`: asserts no `section` field without section arg; `section`+`section_value` present for all known sections; `ok:false`+`error` for unknown section.

Pinpoint: ROADMAP #126

### DIFF
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -6207,7 +6207,7 @@ fn render_config_report(section: Option<&str>) -> Result<String, Box<dyn std::er
 }
 
 fn render_config_json(
-    _section: Option<&str>,
+    section: Option<&str>,
 ) -> Result<serde_json::Value, Box<dyn std::error::Error>> {
     let cwd = env::current_dir()?;
     let loader = ConfigLoader::default_for(&cwd);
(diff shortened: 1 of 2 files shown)


code-yeongyu:
QA note: `resumed_status_command_emits_structured_json_when_requested` fails on this branch but this is a **pre-existing failure on `main`** — not introduced by this PR. The test expects exit 0 but gets exit 1 for `--resume SESSION /status --output-format json`. This is a `main` regression that needs fixing separately before these PRs can show clean CI on that specific test.

PR #3002, opened 2026-05-06
Codex/fix key memory leaks

Rishiatweb:


1716775457damn:
Good catch on the memory leaks. The approach of fixing key retention patterns looks correct. Please make sure tests cover the fixed paths. --name Comment

PR #3014, opened 2026-05-10
Add Qwen model token limits for DashScope compatibility

wangguan1995:
## Summary
  Fixes DashScope compatibility for Qwen models by adding explicit token limits for qwen-max and
  qwen-plus, preventing claw from sending an oversized max_tokens value when routing through the
  OpenAI-compatible provider.

  ## Root Cause
  rust/crates/api/src/providers/mod.rs only defines output token limits for a small set of known
  models.
  When qwen/qwen-max was used, the provider normalized it to qwen-max but did not find a
  matching entry in model_token_limit(), so it fell back to the generic default in
  max_tokens_for_model(). That default was larger than DashScope allows for qwen-max, causing
  the API to reject requests with:

```bash
  400 Bad Request: Range of max_tokens should be [1, 8192]
```
  Changes
  Two small additions in rust/crates/api/src/providers/mod.rs:

  - Add an explicit ModelTokenLimit entry for qwen-max
  - Add an explicit ModelTokenLimit entry for qwen-plus

  This ensures Qwen requests use a provider-specific max_tokens ceiling instead of the generic
  fallback.

  ## Verification

  - cargo build --workspace succeeds
  - ./target/debug/claw --model "qwen/qwen-max" prompt "hello" now succeeds
  - Before the change, the same command failed with a DashScope max_tokens range error

  ## Before / After

  ### Before
  ./target/debug/claw --model "qwen/qwen-max" prompt "hello"
  ✘ Request failed
  error: api returned 400 Bad Request
  InternalError.Algo.InvalidParameter: Range of max_tokens should be [1, 8192]

  ### After
  ./target/debug/claw --model "qwen/qwen-max" prompt "hello"
  ✔ Done
  Hello! How can I assist you with your Rust project today?


code-yeongyu (changes requested):
This PR has merge conflicts (likely from Cargo.lock changes after recent merges). Please rebase onto main and resolve conflicts. Also note the Cargo.lock mode change (100644→100755) appears unintentional.

PR #3017, opened 2026-05-10
feat: interactive provider wizard with fast model selection

TheArchitectit:


1716775457damn:
Interactive provider wizard sounds useful — reducing the friction of model selection is important for onboarding. Could you add a description or screenshots showing the wizard flow? Without seeing the diff it is hard to give detailed feedback, but the concept of fast model selection from a provider list is a good UX direction.

PR #3027, opened 2026-05-14
Convert README to HTML format

icakinser:
Replace README content with HTML structure and styling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,210 +1,592 @@
-# Claw Code
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Claw Code — Local Fork Overview</title>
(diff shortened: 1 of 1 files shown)


code-yeongyu (changes requested):
Converting the README from Markdown to inline HTML makes it harder to read on GitHub (which renders Markdown natively) and breaks GitHub-flavored features like relative links, syntax highlighting, and dark mode compatibility. Please keep the README as Markdown. If you want visual improvements, submit a Markdown-only PR instead.

1716775457damn:
Agree with the requested changes. GitHub natively renders Markdown, so README.md is more maintainable and portable than HTML. Converting would make future edits harder without benefit.

Yeachan-Heo:
Thanks for the effort, but this rewrite is too broad for this repository. It replaces the Markdown README with a full HTML document, embedded styling, external font loading, and a wholesale restructuring of onboarding content. That is unsafe/out-of-scope for a normal docs PR and would make the README harder to maintain. Closing/rejecting this rather than merging it.

PR #3040, opened 2026-05-14
fix: recognize OPENAI_API_KEY as valid auth for OpenAI-compatible endpoints

EmreCelenli:
Previously, claw doctor would warn 'no supported auth env vars were found' even when OPENAI_API_KEY was set, because only ANTHROPIC_API_KEY and ANTHROPIC_AUTH_TOKEN were checked. This caused auth failures when using local models via Ollama or other OpenAI-compatible endpoints.

Tested on Apple M1 MacBook Pro 16GB with Ollama + qwen3:8b.


1716775457damn:
This is a common gotcha — many users setting up OpenAI-compatible endpoints naturally use OPENAI_API_KEY and are confused when it is not recognized. Good interop fix.

PR #3042, opened 2026-05-16
feat(tools): add LoggingAspect to unified tool dispatch entry point

yijunyu:
## Summary

Annotates `execute_tool_with_enforcer` — the single match dispatch every tool
invocation flows through — with `aspect_std::LoggingAspect`, creating a
structured audit trail for tool execution where none currently exists.

Adds three small deps (`aspect-core`, `aspect-macros`, `aspect-std`, all pinned to a git rev) plus `log = "0.4"` to the `tools` crate. No behavioral change.

## Why

A scan of the Rust crates at `a389f8d` turned up zero uses of `tracing::` or `log::`. Every tool invocation, provider call, and hook currently emits either nothing or a raw `println!`/`eprintln!`. For an agent harness that runs `bash`, writes files, and calls third-party LLMs, this is a concrete observability gap.

This PR takes the smallest possible step: one annotation on the one function every tool already passes through. It:

- Creates a uniform entry/exit/error audit log for all 40+ tool branches.
- Uses the `log` facade so downstream consumers pick any subscriber (env_logger, simple_logger, tracing-log) without library-side lock-in.
- Weaves at compile time via the `aspect-rs` proc-macro — no runtime dispatch cost.

## Diff

+12 / −0 lines across 2 files (plus Cargo.lock):

- `rust/crates/tools/Cargo.toml`: +4 dep lines
- `rust/crates/tools/src/lib.rs`: +2 import lines, +1 `#[aspect(...)]` annotation on `execute_tool_with_enforcer`

The aspect-rs deps are pinned to git rev `705a2890` (https://github.com/yijunyu/aspect-rs) rather than crates.io while the API stabilises; a tagged crates.io release will follow once the API has had external review.

## Test plan

- [x] `cargo check -p tools` — clean
- [x] `cargo test -p tools` — same 6 pre-existing failures as `main` (all workspace-permission tests, unrelated to this PR); 100 passing tests unchanged
- [ ] Manual: run the CLI with `RUST_LOG=info` and verify `execute_tool_with_enforcer` entry/exit lines appear for every tool call
- [ ] Manual: run with `RUST_LOG=off` and verify zero log emission (filter fast-path)

## Follow-ups (not in this PR)

- PR #2: Same annotation on `api::ProviderClient::send_message`
- PR #3: Stack `aspect_std::TimingAspect` on both
- PR #4: `aspect_std::RateLimitAspect` on provider calls — closes the rate-limiting gap (producer side currently absent)

Each follow-up is a similar ~10-line PR. This PR intentionally stays narrow to prove the aspect-rs toolchain integrates cleanly with the workspace.

## Fallback if maintainers prefer no external crate

The identical effect in ~30 lines of hand-rolled wrapper code is available; happy to switch on request. The declarative aspect form scales better when the second, third, and fourth aspects stack on the same callsite, which is why I led with it.

## References

- aspect-rs: https://github.com/yijunyu/aspect-rs

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -43,6 +43,35 @@ dependencies = [
 ]
 
 [[package]]
+name = "aspect-core"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70188b9bf884266a6c7117e30af44f38229bc5ac56916bd16512b3e49f90fe20"
+
+[[package]]
+name = "aspect-macros"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "176e7db9b6a7bb4f117b8d97054d2d5a7bdc43b95c19c15c751fb8dcb9bc8a5c"
+dependencies = [
+ "aspect-core",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "aspect-std"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba7d130884fda30ec0acabcadc8f4711267d1cc21edc8f85283e91a011d699fa"
+dependencies = [
+ "aspect-core",
+ "log",
+ "parking_lot",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


1716775457damn:
Good catch on the observability gap — no structured logging across 40+ tool branches is a real blind spot. The aspect-rs approach keeps it zero-cost at runtime while adding entry/exit/error auditing. Using the log facade rather than pulling in a specific subscriber is the right call for library code. Small diff, clean change. LGTM.

Issue #3020, opened 2026-05-12
OpenAI-compatible model IDs with slashes are stripped before request


daekp:
## Summary

When using ClawCode with an OpenAI-compatible local server, model IDs that already contain a slash can fail because ClawCode treats the first path segment as a routing prefix and strips it before sending the request.

In my case, the local server exposes the model as:

```text
openai/gpt-oss-120b
```

But running ClawCode with:

```bash
./target/debug/claw --model openai/gpt-oss-120b prompt "hello"
```

sends this model ID to the backend:

```text
gpt-oss-120b
```

The backend then returns:

```text
✘ ❌ Request failed
[error-kind: api_http_error]
error: api returned 404 Not Found: {"detail":"Not Found"}

Run `claw --help` for usage.
```

## Environment

- ClawCode built from source
- OpenAI-compatible local endpoint
- Backend: OpenAI-compatible local server reporting `owned_by: "vllm"`
- Base URL:

```bash
OPENAI_BASE_URL=http://127.0.0.1:7777/v1
```

The `/v1/models` endpoint returns:

```json
{
  "id": "openai/gpt-oss-120b",
  "owned_by": "vllm"
}
```

## Expected Behavior

ClawCode should allow users to route through the OpenAI-compatible provider while preserving the original model ID sent to the backend.

For example, there should be a way to use:

```bash
--model openai/gpt-oss-120b
```

and have the backend receive:

```text
openai/gpt-oss-120b
```

## Actual Behavior

ClawCode strips the `openai/` prefix and sends:

```text
gpt-oss-120b
```

This breaks local OpenAI-compatible servers whose model IDs include provider-style slugs.

## Possible Solution

Add an opt-in setting or environment variable to preserve routing prefixes for OpenAI-compatible backends.

For example:

```bash
CLAW_OPENAI_PRESERVE_MODEL_PREFIX=1
```

Then `strip_routing_prefix()` could skip removing `openai/`, `qwen/`, etc. when this option is enabled.

A local patch like this worked for me:

```rust
fn strip_routing_prefix(model: &str) -> &str {
    if preserve_routing_prefix_for_wire() {
        return model;
    }

    // existing prefix stripping behavior...
}

fn preserve_routing_prefix_for_wire() -> bool {
    std::env::var("CLAW_OPENAI_PRESERVE_MODEL_PREFIX")
        .map(|value| matches!(value.as_str(), "1" | "true" | "TRUE" | "yes" | "YES" | "on" | "ON"))
[...]

PR #3044, opened 2026-05-17
feat(providers): add "local/" routing prefix for OpenAI-compatible local servers

np6126:
## Summary

A common deployment is running an OpenAI-compatible inference server locally — Ollama, LM Studio, vLLM, llama.cpp — and routing claw-code's OpenAI provider client at it via `OPENAI_BASE_URL`.

The obvious model id formats actively misroute today:

- `qwen/qwen3:14b` or `qwen3-coder` → `metadata_for_model` maps the qwen family to DashScope (`DASHSCOPE_API_KEY`, `dashscope.aliyuncs.com`). A user serving Qwen3 locally on Ollama is silently routed to Alibaba's cloud.
- `kimi/kimi-k2.5` → same DashScope routing.
- `grok-*` → routes to xAI.

The existing `openai/` prefix would work as a workaround, but mentally conflates "this is my local Qwen3" with "the OpenAI API itself", and a user who *does* want the existing OpenRouter behaviour (slug preserved on the wire for `openai/...`) can't have it both ways.

This PR adds `local/` as an explicit routing prefix that says "this is a local OpenAI-compatible server, route as OpenAI client, strip the prefix on the wire". Routing semantics are identical to `openai/`: `OpenAi` provider, `OPENAI_API_KEY` (typically an unused placeholder for local servers), `OPENAI_BASE_URL`.

## Diff

```rust
// providers/mod.rs
-if canonical.starts_with("openai/") || canonical.starts_with("gpt-") {
+if canonical.starts_with("openai/")
+    || canonical.starts_with("local/")
+    || canonical.starts_with("gpt-")
+{
     return Some(ProviderMetadata { provider: ProviderKind::OpenAi, ... });
 }

// providers/openai_compat.rs (wire_model_for_base_url)
-if matches!(lowered_prefix.as_str(), "xai" | "grok" | "qwen" | "kimi") {
+if matches!(lowered_prefix.as_str(), "local" | "xai" | "grok" | "qwen" | "kimi") {
     return Cow::Borrowed(&model[pos + 1..]);
 }
```

Two places only — both pure additions.

**No change to `strip_routing_prefix`** — it's `#[allow(dead_code)]` and only referenced from its own unit tests; adding `local` there would be cosmetic.

**No change to the OpenAI gateway slug preservation logic** used by OpenRouter and similar gateways with a non-default `OPENAI_BASE_URL` — `openai/...` behaviour stays exactly as upstream.

## Example

```bash
export OPENAI_BASE_URL=http://localhost:11434/v1
export OPENAI_API_KEY=ollama-placeholder
claw --model local/qwen3:14b "..."
```

Wire request: `POST $OPENAI_BASE_URL/chat/completions` with `model=qwen3:14b`. No misrouting to DashScope.

## Test plan

- [x] `cargo check --workspace` clean
- [ ] Existing provider tests: `cargo test -p api`
- [ ] Manual: route `local/qwen3:14b` against a local Ollama, verify the request goes to the local server with bare model id on the wire

## Companion discussion

A separate issue (#3045) covers two related open questions: `<think>...</think>` block filtering in streamed output, and whether the `openai/`-slug preservation logic should remain non-configurable once `local/` is available. Posted separately because both involve tradeoffs upstream may want to weigh in on before code lands.

### DIFF
--- a/rust/crates/api/src/providers/mod.rs
+++ b/rust/crates/api/src/providers/mod.rs
@@ -254,7 +254,10 @@ pub fn metadata_for_model(model: &str) -> Option<ProviderMetadata> {
     // route to the correct provider regardless of which auth env vars are set.
     // Without this, detect_provider_kind falls through to the auth-sniffer
     // order and misroutes to Anthropic if ANTHROPIC_API_KEY is present.
-    if canonical.starts_with("openai/") || canonical.starts_with("gpt-") {
+    if canonical.starts_with("openai/")
+        || canonical.starts_with("local/")
+        || canonical.starts_with("gpt-")
+    {
         return Some(ProviderMetadata {
             provider: ProviderKind::OpenAi,
             auth_env: "OPENAI_API_KEY",
--- a/rust/crates/api/src/providers/openai_compat.rs
+++ b/rust/crates/api/src/providers/openai_compat.rs
@@ -939,7 +939,7 @@ fn wire_model_for_base_url<'a>(
         return Cow::Borrowed(&model[pos + 1..]);
     }
 
-    if matches!(lowered_prefix.as_str(), "xai" | "grok" | "qwen" | "kimi") {
+    if matches!(lowered_prefix.as_str(), "local" | "xai" | "grok" | "qwen" | "kimi") {
         return Cow::Borrowed(&model[pos + 1..]);
     }
 


shakoorshkh:
Model id formats like qwen/qwen3:14b silently misroute to DashScope when the user is running a local Ollama server. No error — wrong destination.

Adds a new prefix convention (local/) that users must learn. If OPENAI_BASE_URL is not set, local/ routes to OpenAI with a stripped model id, which will likely 404.


Explicit prefix over inference. Users targeting local servers now carry that intent in the model string rather than relying on routing heuristics to guess correctly.

1148330040:
c5054497-e747-4a0a-9aa4-81ebf9764e2f

Can you help me take a look? I am not familiar with a large model using xinference and have modified the mod.rs and openai_compat.rs files according to your commit.
But when running, it got stuck here directly. I passed in an incorrect model name, and it still remains the same, which proves that it was probably blocked at the step of verifying the apikey and URL. What could be the reason for the error?

1148330040:
> c5054497-e747-4a0a-9aa4-81ebf9764e2f
> Can you help me take a look? I am not familiar with a large model using xinference and have modified the mod.rs and openai_compat.rs files according to your commit. But when running, it got stuck here directly. I passed in an incorrect model name, and it still remains the same, which proves that it was probably blocked at the step of verifying the apikey and URL. What could be the reason for the error?

I suspect that the URL cannot be connected, so I used ngrok to penetrate the local address and tested it. The connection was successful。
793b00d5-5b3d-49ad-8f7f-6713e7818da6

np6126:
> > c5054497-e747-4a0a-9aa4-81ebf9764e2f
> > Can you help me take a look? I am not familiar with a large model using xinference and have modified the mod.rs and openai_compat.rs files according to your commit. But when running, it got stuck here directly. I passed in an incorrect model name, and it still remains the same, which proves that it was probably blocked at the step of verifying the apikey and URL. What could be the reason for the error?
> 
> I suspect that the URL cannot be connected, so I used ngrok to penetrate the local address and tested it. The connection was successful。 793b00d5-5b3d-49ad-8f7f-6713e7818da6

Do you still have the issue? Can you provide a fork with your changes? I'll have a look if you want

1148330040:
> > > c5054497-e747-4a0a-9aa4-81ebf9764e2f
> > > Can you help me take a look? I am not familiar with a large model using xinference and have modified the mod.rs and openai_compat.rs files according to your commit. But when running, it got stuck here directly. I passed in an incorrect model name, and it still remains the same, which proves that it was probably blocked at the step of verifying the apikey and URL. What could be the reason for the error?
> > 
> > 
> > I suspect that the URL cannot be connected, so I used ngrok to penetrate the local address and tested it. The connection was successful。 793b00d5-5b3d-49ad-8f7f-6713e7818da6
> 
> Do you still have the issue? Can you provide a fork with your changes? I'll have a look if you want

Thank you for your enthusiasm, but I have passed the test with two limitations:
1: My local URL here needs to be penetrated
2: Need to ensure that the large model does not have /think

3ea60448-3cca-41e7-a75a-78e1654fc603

1716775457damn:
This is the right fix for #3020 — the local/ prefix cleanly disambiguates routing intent vs upstream model IDs. Good approach: it avoids overloading the slash semantics in the default openai_compat path. Only concern is backward compat — existing users with local servers may need to add the prefix, so a release note callout would help.

niStee:
Bump. This is exactly what's needed to resolve the local Ollama 404 routing bugs reported in #3051. Would love to see this merged!

1716775457damn:
Agreed, the local/ prefix is a clean solution for the Ollama routing confusion described in #3051. The implementation is minimal (2 files) and doesn't break existing behavior. One suggestion: consider adding a brief note in the README or docs about the local/ prefix alongside the existing openai/ documentation so new users discover it naturally.

PR #3049, opened 2026-05-19
feat: add native Gemini support to openai_compat provider

Ajinkya-Kardile:
## Description

This PR enables native integration with Google Gemini by updating the `wire_model_for_base_url` logic in the `openai_compat` provider.

Previously, when using a custom `OPENAI_BASE_URL` (such as the Google Generative Language API endpoint), `claw-code` would preserve the `openai/` prefix in the outgoing JSON payload. This caused API requests to fail because the Google API does not recognize model identifiers like:

```text
models/openai/gemini-....
```

This change adds `gemini` to the stripping logic, ensuring that model slugs are correctly normalized to the format expected by the Google Gemini API while maintaining compatibility for other OpenAI-compatible gateways.

---

## Commands Run

To verify this change, the following commands were executed inside the `rust/` workspace:

```bash
cargo fmt --all --check
```

✅ Passed

```bash
cargo clippy --workspace
```

✅ Passed

```bash
cargo test --workspace
```

✅ Passed

---

### Environment

```bash
export OPENAI_BASE_URL="https://generativelanguage.googleapis.com/v1beta/openai/"
```

### Command Executed

```bash
./target/release/claw --model openai/gemini-2.5-flash prompt "reply with ready"
```

### Result

✅ Successfully received a response from Gemini.

---

### CLI / Provider Behavior

This changes how `openai/` model slugs are sanitized when a non-default `OPENAI_BASE_URL` is detected.

The logic specifically targets models containing `"gemini"` to avoid prefix corruption while preserving compatibility with existing OpenAI-compatible providers.

### Windows / PowerShell

No changes were made to build scripts, shell integrations, or documentation.

The modification is isolated entirely within the Rust `api` crate.

---

## Known Gaps

None.

This implementation follows the existing normalization pattern already used for providers such as:

- `xai`
- `qwen`
- `kimi`

and adheres to the current crate boundaries and provider architecture.


1716775457damn:
Native Gemini support via openai_compat is a clean approach. Stripping the openai/ prefix for the Google API endpoint is the right fix — the previous behavior would never work with Gemini. The manual verification with OPENAI_BASE_URL set to the v1beta/openai endpoint is solid. One question: does this handle the gemini-2.5-flash model slug correctly for both chat and streaming endpoints? Overall looks good.

PR #3058, opened 2026-05-22
Fix: echo reasoning_content back for DeepSeek V4 multi-turn tool calls (fixed the Deepseek V4 thinking problem)

Psy-lzh:
Fix: echo reasoning_content back for DeepSeek V4 multi-turn tool calls
 
Two layers in the CLI silently dropped assistant thinking blocks before they could reach build_chat_completion_request, breaking multi-turn tool-using conversations with DeepSeek V4 (which mandates that prior reasoning_content be echoed back per its thinking-mode protocol).
 
1. AnthropicRuntimeClient::consume_stream did not accumulate 
   ThinkingDelta into pending state, and never emitted
   AssistantEvent::Thinking on ContentBlockStop. As a result the
   session never contained a ContentBlock::Thinking.
 
3. convert_messages mapped ContentBlock::Thinking -> None, dropping
   thinking blocks even if they had been persisted.
 
build_chat_completion_request already serializes
InputContentBlock::Thinking to the reasoning_content field for
DeepSeek V4 (gated on model_requires_reasoning_content_in_history),
so no provider-side changes are needed.
 
Verified end-to-end with deepseek-v4-flash via a local SSE proxy:
multi-turn tool-using conversations now succeed where they
previously failed with "400 The reasoning_content in the thinking
mode must be passed back to the API."

### DIFF
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -8767,6 +8767,8 @@ impl AnthropicRuntimeClient {
         let mut markdown_stream = MarkdownStreamState::default();
         let mut events = Vec::new();
         let mut pending_tool: Option<(String, String, String)> = None;
+        // 累积 reasoning_content 到 Thinking 块（修复 DeepSeek V4 reasoning_content 协议 bug）
+        let mut pending_thinking: Option<(String, Option<String>)> = None;
         let mut block_has_thinking_summary = false;
         let mut saw_stop = false;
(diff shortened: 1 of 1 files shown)


Psy-lzh:
Thanks a lot for the sharing by the way

1716775457damn:
This is an important fix — without echoing reasoning_content back in multi-turn tool calls, DeepSeek V4 loses its chain-of-thought between turns, which severely degrades response quality. The change looks minimal and targeted.

Issue #3062, opened 2026-05-24
error: Context window blocked


LLuke:
During conversation, claw crashed with context_window_blocked

```
... (long output from source code searching by claw)
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum ToolSource {
Base,
… output truncated for display; full result preserved in session. 
✘ ❌ Request failed
[error-kind: unknown]
error: Context window blocked
  Failure class    context_window_blocked
  Session          session-1779586711046-0
  Detail           Requested token count exceeds the model's maximum context length of 196608 tokens. You requested a total of 218999 token…

Recovery
  Compact          /compact
  Resume compact   claw --resume session-1779586711046-0 /compact
  Fresh session    /clear --confirm
  Reduce scope     remove large pasted context/files or ask for a smaller slice 
  Retry            rerun after compacting or reducing the request

Run `claw --help` for usage.
```

1716775457damn:
Context window exhaustion is a real pain point for long-running agent sessions. I see this was resolved and related auto_compact fix (#3106) addresses the root cause by compacting after each iteration rather than only after user turns. Good to see this closed.

PR #3063, opened 2026-05-24
fix: return conservative token limits for unspecified models

LLuke:
to avoid crashes

## Summary
- When working with a model not specially handled and coded in rust/crates/api/src/providers/mod.rs, the claw doesn't have an idea about the context window size. It will keep on growing without automatic compact, eventually lead of crash of the claw process when an error is thrown in the main.rs. The proper way fix is configure and don't hardcode any. However, we still need a reasonable default du jour to reduce chance of crashes. Thus it is picked to be max_output_tokens: 16_384, context_window_tokens: 131_072 as the minimum of each among the current specified models. While this doesn't prevent crash while working with much less capable models and may under-utilize much more capable models, it is a line in the sand currently needed before claw-code further optimize.

## Anti-slop triage
- Evidence:  https://github.com/ultraworkers/claw-code/issues/3062

## Verification
- [X] Targeted tests/docs checks ran, or the gap is explicitly recorded.
- [X] `git diff --check` passes.
- [X] No live secrets, tokens, private logs, or unrelated generated churn are included.

## Resolution gate
- [X] If this PR resolves an issue, the issue number and fix evidence are linked.
- [X] If this PR should not merge, the rejection/defer rationale is evidence-backed and does not rely on vibes.
- [X] I did not merge/close remote PRs or issues from an automation lane without owner approval.

### DIFF
--- a/rust/crates/api/src/providers/mod.rs
+++ b/rust/crates/api/src/providers/mod.rs
@@ -640,7 +640,14 @@ pub fn model_token_limit(model: &str) -> Option<ModelTokenLimit> {
             max_output_tokens: 16_384,
             context_window_tokens: 256_000,
         }),
-        _ => None,
+        // Hotfix: Unknown models get conservative defaults to avoid crashes.
+        // Uses the minimum of known supported models: max_output_tokens: 16_384,
+        // context_window_tokens: 131_072. This may under-utilize the model's
+        // actual capabilities but hopefully ensures safer operation.
+        _ => Some(ModelTokenLimit {
+            max_output_tokens: 16_384,
+            context_window_tokens: 131_072,
+        }),
     }
 }
 


LLuke:
After the fix, testing clawd with MiniMax M2.5 with up to 192K token context window is stable with `auto-compacted` seen:

```
...
[auto-compacted: removed 61 messages]
...
[auto-compacted: removed 8 messages]
```
The 16K/132K were much lower than the model's full capacity but as a safe default the clawd code no longer hangs or crash, -- it is rather snappy.

1716775457damn:
Sensible default — returning 0 or unlimited for unknown models could cause silent truncation or downstream errors. 64K input / 8K output is a reasonable conservative fallback that won't break anything.

PR #3100, opened 2026-05-25
fix(#458): add status to skills and agents JSON envelopes

Yeachan-Heo:
## Summary
- add top-level `status: "ok"` to successful skills JSON envelopes
- add top-level `status: "ok"` to successful agents JSON envelopes
- completes the ROADMAP #458 acceptance sweep for the standard JSON subcommand set

## Verification
- `cargo test --manifest-path rust/Cargo.toml -p commands`
- `cargo build --manifest-path rust/Cargo.toml -p rusty-claude-cli --bin claw`
- `for c in status mcp skills agents doctor sandbox init system-prompt version; do ./rust/target/debug/claw "$c" --output-format json </dev/null | jq -e ' .status | IN("ok","warn","error") ' >/dev/null; done`\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

### DIFF
--- a/rust/crates/commands/src/lib.rs
+++ b/rust/crates/commands/src/lib.rs
@@ -3627,6 +3627,7 @@ fn render_agents_report_json(cwd: &Path, agents: &[AgentSummary]) -> Value {
         "kind": "agents",
         "status": "ok",
         "action": "list",
+        "status": "ok",
         "working_directory": cwd.display().to_string(),
         "count": agents.len(),
         "summary": {
@@ -3710,6 +3711,7 @@ fn render_skills_report_json(skills: &[SkillSummary]) -> Value {
         "kind": "skills",
         "status": "ok",
         "action": "list",
+        "status": "ok",
         "summary": {
             "total": skills.len(),
             "active": active,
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Good PR — the test assertions are exactly what's needed and were missing from main. Two issues to fix before merging:

1. **Duplicate `"status": "ok"` keys** in `render_agents_report_json` and `render_skills_report_json`: main (`0581894b`) already added `"status": "ok"` to both `json!` blocks, and this PR adds a second copy. `serde_json::json!` silently takes the last value for duplicate keys so it's not a runtime bug, but it will likely trigger a clippy warning and is confusing. Please remove the duplicates — the `+` lines at `@@ -3627` and `@@ -3710` are the ones to drop.

2. **`render_skill_install_report_json` `status` key**: main (`cc1462a7`) already added this one too, so the `+        "status": "ok",` at `@@ -3744` will also be a duplicate. The test block for it (`json_report["status"] == "ok"` etc.) is new and valuable — keep the test, drop the duplicate JSON key.

After a rebase onto `cc1462a7` the diff should be **test assertions only** (+11/-0 or similar). That version is clean and mergeable.

PR #3113, opened 2026-05-25
Guard pre-push hook contract against drift

Yeachan-Heo:
## Summary
- add a lightweight regression for the local pre-push hook skip escape hatch
- assert the documented `cargo build --manifest-path rust/Cargo.toml --workspace --locked` gate stays in the hook
- update the G013 verification map to list the regression command

## Verification
- `bash -n .github/hooks/pre-push`
- `python3 tests/test_pre_push_hook_contract.py -v`
- `git diff --check`

## Notes
- The hook implementation in current `main` already has the skip hatch and `--locked`; this PR adds the missing drift regression so it cannot silently regress again.


1716775457damn:
Guarding pre-push hook contract against drift is a good preventative measure — hooks that silently go out of sync with the implementation are a common source of CI-only failures. Tests that validate the doc matches the actual hook behavior are exactly the right approach.

PR #3114, opened 2026-05-25
docs(roadmap): record stale local dogfood probe trap

Yeachan-Heo:
## Summary
- Add ROADMAP #719 for stale local `cargo run` dogfood probes producing authoritative-looking JSON after current main already fixed the behavior.
- Cross-reference sibling provenance/preflight gaps #324 and #695 and fixed plugin-show context #718/#3113.
- Keep scope docs-only with required fix shape for source/build/upstream freshness fields and stale/dirty checkout warnings.

## Verification
- `python3 .github/scripts/check_doc_source_of_truth.py`
- `python3 scripts/validate_cc2_board.py --board .omx/cc2/board.json`
- `git diff --check`


1716775457damn:
Documenting the stale local dogfood probe trap is a good practice. These kinds of silent state corruption issues are hard to reproduce, so having a record of the failure mode helps future debugging. Clean docs-only PR.

PR #3115, opened 2026-05-25
Fix ROADMAP duplicate id guard

Yeachan-Heo:
## Summary
- Add `scripts/roadmap-check-ids.sh` to fail on duplicate helper-era ROADMAP ids, with optional `--min-id` and optional ROADMAP path.
- Wire the guard into docs CI and the repository pre-push hook.
- Document the `roadmap-next-id.sh` + duplicate-check workflow and add ROADMAP #724 as done.

## Anti-slop triage
- Classification: docs-only
- Evidence: `scripts/roadmap-check-ids.sh` passes on current ROADMAP and fails a temp copy with duplicate `723.`.

## Verification
- [x] `bash -n scripts/roadmap-check-ids.sh scripts/roadmap-next-id.sh .github/hooks/pre-push`
- [x] `scripts/roadmap-check-ids.sh`
- [x] temp ROADMAP copy with appended duplicate `723.` failed nonzero and listed duplicate id 723
- [x] `SKIP_CLAW_PRE_PUSH_BUILD=1 .github/hooks/pre-push`
- [x] `git diff --check`
- [x] `python3 .github/scripts/check_doc_source_of_truth.py`
- [x] `python3 .github/scripts/check_release_readiness.py`

## Notes
- Default scope is ids `>=723` because current `main` contains legacy numbered lists and pre-helper duplicate low ids; `--min-id 1` is available for a future strict whole-file audit after historical cleanup.


1716775457damn:
Preventing helper-era ID collisions is a real edge case that could silently produce broken roadmap output. The guard placement before the review loop is the right spot — catches the error early rather than letting corrupted state propagate.

PR #3116, opened 2026-05-26
Fix ROADMAP next-id duplicate awareness

Yeachan-Heo:
## Summary
- make `scripts/roadmap-next-id.sh` run `roadmap-check-ids.sh` from its script directory before printing a next id
- fail closed with a useful stderr error if the checker is unavailable
- add focused stdlib `unittest` coverage for clean output, duplicate fail-fast, and missing-checker fail-closed behavior
- wire the ROADMAP helper behavior test into the existing docs/ROADMAP CI path
- append ROADMAP #725 as DONE using the helper

## Verification
- `scripts/roadmap-next-id.sh ROADMAP.md` printed `725` before appending #725; after appending #725 it prints `726`
- `scripts/roadmap-check-ids.sh ROADMAP.md`
- temp ROADMAP with duplicate `999.` exits nonzero, prints no next id, and lists duplicate id `999`
- `bash -n scripts/roadmap-next-id.sh scripts/roadmap-check-ids.sh`
- `python -m unittest discover -s tests -p test_roadmap_helpers.py`
- `python -m pytest tests/test_roadmap_helpers.py -q`
- `SKIP_CLAW_PRE_PUSH_BUILD=1 bash .github/hooks/pre-push`


1716775457damn:
Preventing poisoned IDs before allocation is the right defense-in-depth approach. Duplicate ID collisions can silently corrupt the roadmap state. Good to see both the detection logic and corresponding test coverage.

PR #3161, opened 2026-05-27
fix(#3129): handle trailing json format for diff errors

### LINKED ISSUE #3129: diff extra args before --output-format json hang with zero stdout
## Summary

`claw diff` extra-argument error paths can hang with zero stdout when `--output-format json` is placed after the unexpected positional/flag argument.

This is adjacent to ROADMAP #801, but distinct: #801 fixed the non-git `diff` JSON envelope. This probe is in a git checkout and hits malformed `diff` invocations.

## Repro

On main `53953a8` in a git checkout:

```bash
timeout --kill-after=1s 8s ./rust/target/debug/claw diff --bogus-flag --output-format json
timeout --kill-after=1s 8s ./rust/target/debug/claw diff does-not-exist --output-format json
timeout --kill-after=1s 8s ./rust/target/debug/claw diff --cached --bogus-flag --output-format json
```

Observed for all three:

- exit `124` from outer timeout
- stdout bytes: `0`
- stderr only contains the deprecation warning from config load

The same pattern reproduced through `cargo run --quiet --manifest-path rust/Cargo.toml --bin claw -- ...`, so it is not just cargo startup cost.

## Expected

Malformed `claw diff` arguments in JSON mode should return promptly with the standard error envelope, including at least:

- `status:"error"` or `kind:"error"`
- `error_kind` such as `unexpected_extra_args` or `unknown_option`
- non-null `hint`
- `message`
- nonzero exit status

## Why this matters

`diff` is a local read-only preflight command used by claws before commit/review. If malformed invocations hang with no JSON envelope, automation cannot distinguish parser misuse from git subprocess stalls or runtime deadlock.

—

Yeachan-Heo:
## Summary
- keep malformed `claw diff ... --output-format json` invocations on the parser error path
- add regression coverage for trailing JSON output-format after malformed diff args
- preserve nonzero JSON error envelopes with `error_kind`, `hint`, and `message`

Fixes #3129.

## Tests
- `git diff --check`
- `cargo fmt --check`
- `cargo test -p rusty-claude-cli diff_extra_args_have_typed_error_kind_and_hint_766 --test output_format_contract`
- `cargo test -p rusty-claude-cli diff_trailing_json_after_malformed_args_is_bounded_json_3129 --test output_format_contract`
- `cargo test -p rusty-claude-cli diff_non_git_dir_has_error_kind_and_hint_801 --test output_format_contract`

## Notes
- Direct `./rust/target/debug/claw ...` smoke was inconclusive in this environment because local auth/provider configuration intercepted the invocation with an Invalid bearer token response; focused Rust integration tests cover the parser/output contract.

### DIFF
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -1225,10 +1225,10 @@ fn parse_args(args: &[String]) -> Result<CliAction, String> {
         // `git diff`). No session needed to inspect the working tree.
         "diff" => {
             if rest.len() > 1 {
-                return Err(format!(
-                    "unexpected extra arguments after `claw diff`: {}\nUsage: claw diff",
-                    rest[1..].join(" ")
-                ));
+                // #3129: keep malformed `diff ... --output-format json` on the
+                // parser/error path, not the prompt/TUI fallback. The newline
+                // before Usage is part of the JSON hint contract.
+                return Err(unexpected_diff_args_error(&rest[1..]));
             }
             Ok(CliAction::Diff { output_format })
         }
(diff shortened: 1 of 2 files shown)


1716775457damn:
Thanks for this fix. Trailing JSON is a common edge case when CLI tools mix stdout and stderr or append diagnostic output after valid JSON. Good catch.

Issue #3036, opened 2026-05-14
docs: add official Ollama/llama.cpp/vLLM local model examples


Yeachan-Heo:
## Source
Discord `#claw-code` support: users asked whether Claw Code can run all-local with Ollama/local Qwen/llama.cpp and how to configure a non-Anthropic provider.

## Problem
Community guides exist, but local/OpenAI-compatible setup should have official known-good examples and warnings.

## Suggested scope
Add docs for local OpenAI-compatible providers:
- Ollama
- llama.cpp server
- vLLM or other OpenAI-compatible servers

Include examples such as:
```bash
export OPENAI_BASE_URL="http://127.0.0.1:11434/v1"
export OPENAI_API_KEY="dummy"
claw --model "openai/qwen3:latest" prompt "Reply exactly HELLO_WORLD_123"
```

Explain:
- use `openai/` prefix for OpenAI-compatible route
- model id must match what the local server exposes
- raw `/v1/chat/completions` smoke test
- tool-call support may be required for slash/tool workflows
- local routes may be rougher than first-party Anthropic/OpenAI

## Acceptance criteria
- Users can configure at least one local provider from official docs.
- Docs include a one-shot prompt smoke test.
- Docs warn about tool-call/response-shape limitations.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Closing as part of issue tracker triage. This is tracked in ROADMAP.md, is a known limitation, or is off-topic.

Issue #3123, opened 2026-05-26
Ollama API compatibility issue for model name (model not found + wrong provider) 2


Noobzik:
Reopening as a new issue  #3051 
Prior related work : #3044 

> Feedback from the new build, not fixed as it seems [@code-yeongyu](https://github.com/code-yeongyu) [@niStee](https://github.com/niStee) :
> 
> ```
> PS C:\Users\XXX\Documents\git\claw-code\rust> git pull
> remote: Enumerating objects: 5, done.
> remote: Counting objects: 100% (5/5), done.
> remote: Compressing objects: 100% (1/1), done.
> remote: Total 3 (delta 2), reused 3 (delta 2), pack-reused 0 (from 0)
> Unpacking objects: 100% (3/3), 1.13 KiB | 27.00 KiB/s, done.
> From https://github.com/ultraworkers/claw-code
>    ee24ff2..21a9860  main       -> origin/main
> Updating ee24ff2..21a9860
> Fast-forward
>  ROADMAP.md | 2 ++
>  1 file changed, 2 insertions(+)
> PS C:\Users\XXX\Documents\git\claw-code\rust> git checkout main
> Already on 'main'
> Your branch is up to date with 'origin/main'.
> PS C:\Users\XXX\Documents\git\claw-code\rust> cargo build --workspace
> warning: unused import: `std::path::Path`
>  --> crates\plugins\src\hooks.rs:2:5
>   |
> 2 | use std::path::Path;
>   |     ^^^^^^^^^^^^^^^
>   |
>   = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default
> 
> warning: `plugins` (lib) generated 1 warning (run `cargo fix --lib -p plugins` to apply 1 suggestion)
> warning: variable does not need to be mutable
>    --> crates\runtime\src\hooks.rs:740:9
>     |
> 740 |     let mut command_builder = {
>     |         ----^^^^^^^^^^^^^^^
>     |         |
>     |         help: remove this `mut`
>     |
>     = note: `#[warn(unused_mut)]` (part of `#[warn(unused)]`) on by default
> 
> warning: function `save_user_provider_settings` is never used
>    --> crates\runtime\src\config.rs:617:8
>     |
> 617 | pub fn save_user_provider_settings(
>     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |
>     = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default
[...]

PR #3165, opened 2026-05-27
feat: add local/ routing prefix for model IDs with embedded slashes

petterreinholdtsen:
## Summary

Some backends (e.g., gpt.uio.no running Ollama) have model IDs containing slashes like Qwen/Qwen3.6-27B-FP8. The existing routing prefix logic would strip 'Qwen/' as a provider prefix, sending only the remainder on the wire and causing 403 Forbidden from the server.

The new local/ prefix is an escape hatch: it strips just 'local/' and sends everything after it verbatim, preserving embedded slashes in model IDs.

Usage: --model local/Qwen/Qwen3.6-27B-FP8

This patch was developed with assistance from OpenCode and local Qwen 3.6 API server.

## Anti-slop triage

I do not undersnad these fields:

- Classification: 
- Evidence: 

## Verification

I do not understand the two first of these check list points.

## Resolution gate

I do not understand these check list points:

I believe this is related to https://github.com/ultraworkers/claw-code/issues/3036, but unsure if it fix it.

### DIFF
--- a/rust/crates/api/src/providers/openai_compat.rs
+++ b/rust/crates/api/src/providers/openai_compat.rs
@@ -915,14 +915,17 @@ pub fn model_requires_reasoning_content_in_history(model: &str) -> bool {
 
 /// Strip routing prefix (e.g., "openai/gpt-4" → "gpt-4") for the wire.
 /// The prefix is used only to select transport; the backend expects the
-/// bare model id.
+/// bare model id. Use `"local/"` as an escape hatch when the server's model
+/// ID itself contains a slash (e.g., `local/Qwen/Qwen3.6-27B-FP8`).
 #[allow(dead_code)]
 fn strip_routing_prefix(model: &str) -> &str {
     if let Some(pos) = model.find('/') {
         let prefix = &model[..pos];
         // Only strip if the prefix before "/" is a known routing prefix,
         // not if "/" appears in the middle of the model name for other reasons.
-        if matches!(prefix, "openai" | "xai" | "grok" | "qwen" | "kimi") {
+        // "local" is an escape hatch: strips just "local/" and sends everything
+        // after it verbatim (including embedded slashes) on the wire.
+        if matches!(prefix, "openai" | "xai" | "grok" | "qwen" | "kimi" | "local") {
             &model[pos + 1..]
         } else {
             model
@@ -962,6 +965,12 @@ fn wire_model_for_base_url<'a>(
     }
 
     if matches!(lowered_prefix.as_str(), "xai" | "grok" | "qwen" | "kimi") {
+        return Cow::Borrowed(&model[pos + 1..]);
+    }
+
+    // "local/" escape hatch: strips just the prefix and sends everything after it
+    // verbatim on the wire (including embedded slashes in model IDs).
+    if lowered_prefix == "local" {
         return Cow::Borrowed(&model[pos + 1..]);
     }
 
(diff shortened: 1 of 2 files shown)


petterreinholdtsen:
I asked opencode to keep notes on what it was solving, and here is what it ended up with for this patch.  I've removed secrets and private DNS names.

## Crash: "Access denied to model" on corporate API server (routing prefix stripping) [SOLVED]
- **Error**: `api returned 403 Forbidden: Access denied to model: Qwen3.6-27B-FP8`
- **Root cause**: Corporate API server has model IDs containing slashes like `Qwen/Qwen3.6-27B-FP8`. The existing routing prefix logic in `wire_model_for_base_url()` stripped `Qwen/` as a known provider prefix, sending only `Qwen3.6-27B-FP8` on the wire — which the server doesn't recognize (403). Additionally, `validate_model_syntax()` rejected 3+ parts when split by `/`, so even if you tried to work around it, the CLI would refuse the model name.
- **Key files**:
  - `rust/crates/api/src/providers/openai_compat.rs:920-933` — `strip_routing_prefix()` now accepts `"local"` as a routing prefix
  - `rust/crates/api/src/providers/openai_compat.rs:964-975` — `wire_model_for_base_url()` handles `"local/"` escape hatch
  - `rust/crates/rusty-claude-cli/src/main.rs:1654-1658` — `validate_model_syntax()` now allows embedded slashes for `local/` prefix

### What was done
- Added `"local"` to the routing prefixes in both `strip_routing_prefix()` and `wire_model_for_base_url()`. The `local/` prefix is an escape hatch: it strips just `local/` and sends everything after it verbatim on the wire, preserving embedded slashes in model IDs.
- Updated `validate_model_syntax()` to allow more than 2 parts when split by `/` for models starting with `local/`, so `local/Qwen/Qwen3.6-27B-FP8` passes validation.
- Use: `--model local/Qwen/Qwen3.6-27B-FP8` sends the full ID on the wire.

### Test command
```bash
echo "Say hello." | OPENAI_BASE_URL=https://corporate.example.com/api/v1/chat/completions OPENAI_API_KEY=secret ~/src/claw-code-upstream/rust/target/debug/claw --model local/Qwen/Qwen3.6-27B-FP8
```

1716775457damn:
The local/ escape hatch is a clean solution for model IDs with embedded slashes. Much better than trying to guess which / is a provider separator vs part of the model name. This should also help with the broader Ollama compatibility issues in #3123.

rust/crates/api/src/providers/openai_compat.rs
```diff
@@ -2730,4 +2739,24 @@ mod tests {
...
+
+    #[test]
+    fn strip_routing_prefix_preserves_embedded_slash_with_local_prefix() {
+        // "local/" escape hatch: strips just the prefix, preserves everything after it verbatim
+        assert_eq!(
+            super::strip_routing_prefix("local/Qwen/Qwen3.6-27B-FP8"),
+            "Qwen/Qwen3.6-27B-FP8"
+        );
+        assert_eq!(super::strip_routing_prefix("local/mistralai/Mistral-Large-3"), "mistralai/Mistral-Large-3");
+    }
```
Noobzik, line 2751: I think there should be one more case where you have `local/some_random_model:7b` that should assert to `local/some_random_model:7b` 

Overwise, that look's good for me

Noobzik (commented):
One more test case then it should be good for the maintainer to approve
@code-yeongyu

1716775457damn:
Nice approach. The local/ prefix convention is clean and doesn't conflict with existing provider routing. One thought: since the model ID after local/ is used as-is, servers with query-string authentication (e.g. http://localhost:1234/v1?key=xxx) could have their key leaked into model_id. Consider either stripping query params, or documenting that local/ expects just the host:port and paths, not query strings.

Yeachan-Heo:
Applied the compatible `local/` escape-hatch fix in bcc5bfd (fix: route local OpenAI-compatible models). Main now accepts `--model local/Qwen/Qwen3.6-27B-FP8`, routes it to the OpenAI-compatible provider, strips only the `local/` prefix on the wire, and documents the behavior. Verification passed: `cargo test --manifest-path rust/Cargo.toml -p api local_routing_prefix_strips_only_escape_hatch -- --nocapture`, `local_prefix_routes_to_openai_not_anthropic`, `cargo test --manifest-path rust/Cargo.toml -p rusty-claude-cli parses_single_word_command_aliases_without_falling_back_to_prompt_mode -- --nocapture`, `cargo fmt --manifest-path rust/Cargo.toml --all -- --check`, and `cargo build --manifest-path rust/Cargo.toml --workspace --locked`. Closing this PR as applied/superseded by the focused main-branch implementation.

PR #3185, opened 2026-05-28
Test doctor help JSON flag order

Yeachan-Heo:
## Summary
- extends the doctor help JSON regression to cover suffix flag order (`doctor --help --output-format json`)
- covers topic-help routing (`help doctor --output-format json`) with the same structured local-only contract
- keeps this as a regression-only follow-up after #3184 so the fast path stays machine-discoverable across common flag orders

## Validation
- `cargo test --manifest-path rust/Cargo.toml -p rusty-claude-cli --test output_format_contract doctor_help_json -- --nocapture`
- `cargo build -q --bin claw`
- `./target/debug/claw --output-format json doctor --help`
- `./target/debug/claw doctor --help --output-format json`
- `./target/debug/claw help doctor --output-format json`

—


1716775457damn:
Good catch on flag order edge cases — --help --output-format json and help doctor --output-format json are both common invocation patterns. Regression coverage here makes sense.

PR #3186, opened 2026-05-28
Record argv-safe dogfood probe gap

Yeachan-Heo:
## Summary
- records ROADMAP #813 for the dogfood-harness failure mode where shell-string command loops can fabricate argv-sensitive CLI failures
- documents the fix shape: argv-safe probe helper/recipe, separated stdout/stderr, and probe_error vs product_error labeling
- keeps #3184/#3185 product behavior intact; this is a follow-up on the verification harness/logging surface

## Validation
- docs-only ROADMAP update
- explicit argv recheck on main passed for the three doctor help JSON forms before filing

—


1716775457damn:
Good ROADMAP entry — the argv-safe probe helper pattern is a sensible improvement for dogfood reliability. Clean scope.

PR #3187, opened 2026-05-28
Record trailing JSON inventory timeout gap

Yeachan-Heo:
## Summary
- records ROADMAP #814 from fresh argv-safe dogfood after #3186 merged
- pinpoints plain non-TTY trailing `--output-format json` timeout on inventory/error surfaces
- scopes required follow-up around parsing trailing JSON flag before REPL/provider startup and adding timeout regression coverage

## Dogfood evidence
Fresh `main` `0e6d48d9d` explicit-argv probes:
- `cargo run -q --bin claw -- agents list --bogus --output-format json` -> timeout rc 124, stdout 0 bytes
- `cargo run -q --bin claw -- skills show does-not-exist --output-format json` -> timeout rc 124, stdout 0 bytes
- `cargo run -q --bin claw -- plugins show does-not-exist --output-format json` -> timeout rc 124, stdout 0 bytes

## Validation
- docs-only ROADMAP update

—


1716775457damn:
Good record — the trailing flag timeout is a subtle but real issue for non-interactive JSON usage. Parsing flags before provider startup is the right direction.

PR #3199, opened 2026-05-29
fix: unknown single-word subcommand emits command_not_found instead of missing_credentials (#825)

code-yeongyu:
Fixes ROADMAP #825.

**Before:** `claw foobar` → `missing_credentials` (after full provider startup attempt)
**After:** `claw foobar` → `command_not_found` (immediate, no provider startup)

**Root cause:** `looks_like_subcommand_typo` fired for all single-word all-alpha inputs, but only returned an error when fuzzy suggestions existed. No-suggestion fallthrough reached `CliAction::Prompt` → Anthropic startup → misleading error.

**Fix:** Always emit `command_not_found:` from the typo guard. Added `command_not_found` to the classifier. Unified old `unknown_subcommand` kind under `command_not_found`.

**Tests:** 3 new regression tests, 1 existing test updated, 572 pass.


1716775457damn:
Thorough fix. Unifying old unknown_subcommand under command_not_found and making the typo guard always fire is the right approach. 3 new regression tests give confidence.

PR #3200, opened 2026-05-29
fix: multi-word unknown subcommand emits command_not_found (#826)

code-yeongyu:
Fixes ROADMAP #826.

**Before:** `claw foobar baz` → `missing_credentials` (after provider startup)
**After:** `claw foobar baz` → `command_not_found` (immediate, no provider startup)

**Root cause:** #825 guard only fired when `rest.len() == 1`. Multi-word cases slipped through to `CliAction::Prompt`.

**Fix:** Removed the `rest.len() == 1` restriction from the `looks_like_subcommand_typo` guard. Any first-word that looks like a command word but is unknown now immediately returns `command_not_found`.

**Tests:** 1 new regression test, 572 pass.


1716775457damn:
Good catch — the 
est.len() == 1 guard was exactly the hole that let multi-word cases through. Clean logic.

PR #3201, opened 2026-05-29
fix: resume unknown slash command emits unknown_slash_command error_kind (#827)

code-yeongyu:
Fixes ROADMAP #827.

**Before:** `claw --resume latest --output-format json /bogus` → `error_kind:"unknown"` (opaque)
**After:** `claw --resume latest --output-format json /bogus` → `error_kind:"unknown_slash_command"` (typed)

**Root cause:** `format_unknown_slash_command()` emitted a message without a classifier-friendly prefix, so `classify_error_kind` fell through to the generic `"unknown"` fallback.

**Fix:** Prefix message with `unknown_slash_command:` and add classifier arm.

**Tests:** 1 new regression test, 572 pass.


1716775457damn:
Clean fix. The prefix-based classifier arm is a minimal, targeted approach and the regression test covers the exact case well.

PR #3202, opened 2026-05-29
fix: /approve and /deny outside REPL emit interactive_only error_kind (#828)

code-yeongyu:
Fixes ROADMAP #828.

**Before:** `claw --output-format json /approve` → `error_kind:"unknown_slash_command"`
**After:** `claw --output-format json /approve` → `error_kind:"interactive_only"`

**Root cause:** `/approve`, `/yes`, `/deny`, `/no` are not `SlashCommand` enum variants, so they fell through to `format_unknown_direct_slash_command`. They're valid REPL-only commands that require an active tool call.

**Fix:** Intercept them in `SlashCommand::Unknown` arm with a dedicated `interactive_only:` message.

**Tests:** 1 new test covering all 4 aliases. 572 pass.


1716775457damn:
Makes sense — /approve, /deny, /yes, /no are valid REPL-only commands that shouldn't fall through to unknown_slash_command. The dedicated interactive_only interception covers all 4 aliases cleanly.

PR #3203, opened 2026-05-29
fix: interactive_only hint omits --resume for non-resume-safe commands (#829)

code-yeongyu:
Fixes ROADMAP #829.

**Before:** `claw --output-format json /commit` hint → `"use claw --resume SESSION.jsonl /commit"` (wrong — /commit is not resume-safe)
**After:** `claw --output-format json /commit` hint → `"Start claw and run it there."`

Resume-safe commands like /diff still suggest `--resume` in the hint.

**Root cause:** Generic interactive_only format string always appended `--resume` regardless of whether the command appears in `resume_supported_slash_commands()`.

**Tests:** 2 new tests. 572 pass.

### DIFF
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -1755,12 +1755,26 @@ fn parse_direct_slash_cli_action(
         }
         Ok(Some(command)) => Err({
             let _ = command;
-            format!(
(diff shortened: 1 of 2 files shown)


1716775457damn:
Nice fix — the generic interactive_only hint was misleading for non-resume-safe commands. Separating the hint logic based on resume_supported_slash_commands() is the right approach. Good to see this landed.

PR #3205, opened 2026-05-29
fix: direct-slash resume-safe commands route to CliAction instead of interactive_only (#831)

code-yeongyu:
Fixes ROADMAP #831.

**Before:**
- `claw /status` → `error_kind:"interactive_only"`
- `claw /diff` → `error_kind:"interactive_only"`
- `claw /version` → `error_kind:"interactive_only"`
- `claw /doctor` → `error_kind:"interactive_only"`
- `claw /sandbox` → `error_kind:"interactive_only"`

**After:** Same output as `claw status` / `claw diff` / `claw version` etc.

**Root cause:** `parse_direct_slash_command` only had explicit arms for `/help`, `/agents`, `/mcp`, `/skills`. All other SlashCommand variants fell to the generic `interactive_only` arm.

**Tests:** 1 new integration test, 572+ pass.


1716775457damn:
Thanks for the fix! This properly routes resume-safe commands to CliAction instead of incorrectly emitting interactive_only. The before/after behavior change is clear and the root cause makes sense. One concern: there are 5 failing CI checks (build, clippy, fmt, test, windows smoke) that need to be addressed before merging.

Yeachan-Heo:
Closing as superseded by current main. The duplicate #830 commit in this PR was already implemented, and the remaining direct-slash resume-safe routing work (#831) is now implemented and pushed in `55da189` with focused output-format contract coverage. The PR branch is conflicting and no longer has unique mergeable work.

PR #3209, opened 2026-05-31
Harden permission enforcement against sandbox bypasses

Sam0urr:
Close two ways the permission system could be bypassed:

- Workspace path traversal: normalize `.`/`..` lexically before the
  boundary prefix comparison so paths like `/workspace/../../etc` can no
  longer escape the sandbox. Fixed in both the runtime enforcer and the
  duplicate check in the tools PowerShell path classifier.
- read-only mode no longer trusts the leading token alone: reject shell
  metacharacters (chaining/substitution/redirect/pipe/subshell), drop
  interpreters and build drivers (python/node/ruby/cargo/rustc) from the
  allow-list, gate `git` to non-mutating subcommands, and reject `find`
  actions that execute or delete.

Adds regression tests for both holes. The pre-existing, unrelated
worker_boot git-metadata test failure is not affected by this change.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/rust/crates/runtime/src/permission_enforcer.rs
+++ b/rust/crates/runtime/src/permission_enforcer.rs
@@ -173,32 +173,118 @@ impl PermissionEnforcer {
     }
 }
 
-/// Simple workspace boundary check via string prefix.
+/// Workspace boundary check.
+///
+/// Resolves `.` and `..` components lexically *before* comparing against the
+/// workspace root, so that traversal sequences like `/workspace/../../etc`
+/// cannot escape the sandbox via a naive string prefix match. Normalization is
+/// lexical (it does not touch the filesystem) because the target path may not
+/// exist yet on a write, and we must not depend on CWD.
 fn is_within_workspace(path: &str, workspace_root: &str) -> bool {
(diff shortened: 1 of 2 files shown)


Sam0urr:
Note on CI checks (pre-empting baseline questions):

Two pre-existing issues on main are unrelated to this change, in case they surface in CI:

- Clippy — `cargo clippy --workspace --all-targets -- -D warnings` is already red on main (13 errors in [crates/runtime/src/trident.rs](url) and [crates/runtime/src/config.rs](url), mostly `needless_borrows_for_generic_args`). None are introduced here — `permission_enforcer.rs` and the touched section of [tools/src/lib.rs](url) are clippy-clean.

- Tests — `cargo test --workspace` has one pre-existing failure: `worker_boot::tests::startup_preflight_warns_when_git_metadata_is_not_writable` (environment-dependent git-metadata writability check). I confirmed it fails identically with my changes stashed, so it's not caused by this PR. Everything else passes (578 tests), including the 8 new regression tests added here.

Happy to rebase or split things differently if that's preferred

PR #3213, opened 2026-06-02
Feat/ollama provider

Aneesh-382005:
## Summary

The codebase already acknowledged this gap in a comment in `rust/crates/api/src/providers/mod.rs`:

```rust
// When OPENAI_BASE_URL is set, the user explicitly configured an
// OpenAI-compatible endpoint. Prefer it over the Anthropic fallback
// even when the model name has no recognized prefix — this is the
// common case for local providers (Ollama, LM Studio, vLLM, etc.)
// where model names like "qwen2.5-coder:7b" don't match any prefix.
```

The routing logic beneath it was correct. But two layers upstream were
still blocking local model names before routing ever ran.

**Before this PR**, the intended workaround was:
```bash
export OPENAI_BASE_URL="http://127.0.0.1:11434/v1"
export OPENAI_API_KEY="anything"
claw --model "openai/qwen2.5-coder:7b" prompt "say hello"
```
This required three env vars, a fake API key, and a `provider/` prefix
that Ollama doesn't understand, and it still broke due to a logic bug in
`wire_model_for_base_url` where an inner `matches!` could never fire,
causing the full `openai/qwen2.5-coder:7b` slug to reach Ollama's API
instead of the bare model name.

**After this PR**:
```bash
export OLLAMA_HOST="http://127.0.0.1:11434"
claw --model "qwen2.5-coder:7b" prompt "say hello"
```

One env var. No prefix. No fake key. Matches Ollama's own convention.

**What was blocking it**

Two separate layers rejected local model names before routing logic ran:

1. `validate_model_syntax()` in `rusty-claude-cli/src/main.rs` enforced
   `provider/model` format on every model name unconditionally.

2. `detect_provider_kind()` in `providers/mod.rs` matched the `qwen`
   prefix and routed to DashScope regardless of what endpoint was configured.

**What this PR changes**

- `rusty-claude-cli/src/main.rs` - if `OLLAMA_HOST` is set, skip the
  `provider/model` format requirement. Empty strings still rejected.
- `providers/mod.rs` - if `OLLAMA_HOST` is set, skip prefix matching
  and route directly to the OpenAI-compatible backend.
- `providers/openai_compat.rs` - adds `OLLAMA_CONFIG` constant and
  `from_ollama_env()` constructor that reads `OLLAMA_HOST`, appends `/v1`,
  and sets a dummy api_key since Ollama requires no auth.
- `client.rs` - in the `OpenAi` dispatch arm, when `OLLAMA_HOST` is set,
  use `from_ollama_env()` instead of the standard constructor.
- `docs/local-openai-compatible-providers.md` and `USAGE.md` updated to
  document `OLLAMA_HOST` as the preferred Ollama path.

**Follow-up work not in this PR**

- `ProviderKind::Ollama` as a dedicated variant so `/doctor` shows
  "Ollama" as the provider name rather than "OpenAI"
- Same pattern for LM Studio and llama.cpp
- Tool-call graceful degradation for models that don't support
  function calling

## Anti-slop triage
- Evidence: `claw --model "qwen2.5-coder:7b" prompt "x"` returns
  `invalid model syntax` on unpatched main. With this patch and
  `OLLAMA_HOST=http://127.0.0.1:11434`, `qwen2.5-coder:7b` responds
  correctly. Regression test added in `alias_resolution_tests`.

## Verification
- [x] Targeted tests/docs checks ran, or the gap is explicitly recorded.
- [x] `git diff --check` passes.

## Resolution gate
- [x] If this PR should not merge, the rejection/defer rationale is evidence-backed and does not rely on vibes.
- [x] I did not merge/close remote PRs or issues from an automation lane without owner approval.


Yeachan-Heo:
Thanks for the contribution! The Ollama provider approach is solid — using `OLLAMA_HOST` to route and bypass colon-containing model names makes sense.

However, the diff in `crates/api/src/providers/mod.rs` introduces a broken merge: the added `if let Some(metadata) = metadata_for_model(model)` line creates a dangling unclosed block right before the existing `let resolved_model = ...` code. This won't compile.

Could you rebase this on the latest main and fix the conflict in `detect_provider_kind`? The OLLAMA_HOST early-return should come before the existing `let resolved_model` line, and the original `if let Some(metadata) = metadata_for_model(&resolved_model)` should remain intact after it.

Once the diff is clean I'm happy to merge.

Yeachan-Heo:
Superseded by commit be8112f which adds native Ollama support via  env var. This PR's approach was sound but had a duplicate  bug in . The merged implementation uses  as the preferred single env var, bypasses model syntax validation for Ollama names, and requires no API key.

Aneesh-382005:
Thanks for reviewing and cleaning this up. I'll follow up with generic local provider support for all sorts of frameworks. 
LM Studio, llama.cpp, and vLLM, so bare model names work across all compatible local servers, not just Ollama

1716775457damn:
Nice work on the Ollama integration. The approach of using OLLAMA_HOST to bypass colon-based model name parsing is clean. For the follow-up on generic local provider support across LM Studio/llama.cpp/vLLM, would it make sense to also consider a config-based provider registry so users can add custom endpoints without code changes? That would make the system extensible beyond the built-in providers.

Aneesh-382005:
> Nice work on the Ollama integration. The approach of using OLLAMA_HOST to bypass colon-based model name parsing is clean. For the follow-up on generic local provider support across LM Studio/llama.cpp/vLLM, would it make sense to also consider a config-based provider registry so users can add custom endpoints without code changes? That would make the system extensible beyond the built-in providers.

That's the direction I am thinking as well. A config based registry makes more sense than handling different Local LLM frameworks as edge cases. I'm planning to work on this.

1716775457damn:
Nice analysis of the double-blocking issue. The OLLAMA_HOST env var is a clean approach — it matches how the Ollama ecosystem already works and avoids forcing users through the provider/model syntax that has no meaning in local deployment.

One edge case worth addressing in follow-up: what happens when both OLLAMA_HOST and OPENAI_BASE_URL are set? The current implementation seems to prefer OLLAMA_HOST, but users running mixed local+remote setups might want to differentiate. Consider adding a clear precedence rule in docs, or even an explicit conflict warning if both are present.

Also, for the follow-up work on ProviderKind::Ollama — it would be useful to expose provider-specific capabilities (like whether tool calling is supported) so that /doctor can warn users upfront rather than hitting runtime errors.

PR #3228, opened 2026-06-05
feat: local Ollama support — surface Qwen3 reasoning, model hint, WSL guide

olabi:
## Summary

Make `claw` usable against a local [Ollama](https://ollama.com) server end-to-end, with a focus on Qwen3 "thinking" models:

1. **`fix(api)`** — surface Ollama's `reasoning` field so Qwen3 thinking models don't produce empty output.
2. **`fix(cli)`** — extend the invalid-model hint for `qwen*` to mention the local Ollama path.
3. **`docs`** — add a WSL + Ollama setup guide to `rust/README.md`.

## Motivation

Ollama's Qwen3 thinking models stream the chain-of-thought in a `reasoning` field with an **empty `content`**:

```json
{"choices":[{"message":{"role":"assistant","content":"","reasoning":"Thinking Process: ..."}}]}
```

The OpenAI-compatible provider only recognised `reasoning_content` / `thinking`, so `reasoning`-only responses surfaced **no text at all** ("assistant stream produced no content"). This adds `reasoning` to both the streaming (`ChunkDelta`) and non-streaming (`ChatMessage`) shapes and falls back to it when extracting thinking — **preserving the existing thinking/text content-block separation** (no regression to the GLM/DeepSeek `reasoning_content` path).

## Changes

- `crates/api/src/providers/openai_compat.rs` — recognise the `reasoning` field on both response shapes; fall back to it for thinking extraction; **two new tests** (streaming + non-streaming).
- `crates/rusty-claude-cli/src/main.rs` — invalid-model hint for `qwen*` now also points at the local Ollama path (`OPENAI_BASE_URL` + `openai/` prefix).
- `rust/README.md` — "Local models via Ollama (WSL)" section: build, env vars, the exact colon-tag model name requirement (`qwen3.5:9b`, **not** `qwen3.5-9b`), tag listing, WSL connectivity note, reasoning-model behaviour.

## Verification

Built `--release` and ran against a local Ollama `qwen3.5:9b`:

```
$ export OPENAI_BASE_URL="http://127.0.0.1:11434/v1"
$ export OPENAI_API_KEY="ollama"
$ ./target/release/claw --model openai/qwen3.5:9b "What is 17 + 26? Reply with the number only."
▶ Thinking (collapsed)
43
✔ Done
```

Without this change the same prompt yields an empty assistant turn whenever the model routes its tokens through `reasoning`.

## Test plan

- [x] `cargo test -p api` — new `*_ollama_reasoning_field_*` tests pass, existing `reasoning_content` tests unaffected
- [x] `cargo fmt -p api --check`
- [x] `cargo clippy -p api` (no new warnings)
- [x] Manual end-to-end run against local Ollama `qwen3.5:9b`
- [ ] CI should pass (fmt, clippy, test)

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -63,9 +63,12 @@ export OPENAI_API_KEY="ollama"   # any non-empty value works; Ollama ignores it
 
 **The model name must match the Ollama tag exactly.** Ollama tags use a colon
 (`name:tag`), e.g. `qwen3.5:9b` — not a hyphen (`qwen3.5-9b`), which the server
-rejects with `404 ... model not found`. Routing to the OpenAI-compatible
-provider only kicks in when `OPENAI_BASE_URL` is set and the model name contains
-a `:` or `.`, so always pass the full tag. List the installed tags with:
+rejects with `404 ... model not found`. In the command above the `openai/`
+prefix is what selects the OpenAI-compatible provider (the prefix is stripped
+before the tag is sent to Ollama). You can also drop the prefix and pass the
+bare tag — when `OPENAI_BASE_URL` is set, claw routes a prefix-less model to the
+OpenAI-compatible endpoint as long as the name contains a `:` or `.`. Either
+way, always pass the full colon tag. List the installed tags with:
 
 ```bash
 ollama list                                 # native view
--- a/rust/README.md
+++ b/rust/README.md
@@ -40,6 +40,48 @@ Or provide an OAuth bearer token directly:
 export ANTHROPIC_AUTH_TOKEN="anthropic-oauth-or-proxy-bearer-token"
 ```
 
+### Local models via Ollama (WSL)
+
+Claw works with any OpenAI-compatible endpoint, including a local
+[Ollama](https://ollama.com) server. This is handy for offline use or for
+driving the CLI from WSL against models served on the Windows host.
+
+```bash
+cd rust/
+
+# 1. Build the release binary
+cargo build --release
+
+# 2. Point claw at the local Ollama OpenAI-compatible endpoint
+export OPENAI_BASE_URL="http://127.0.0.1:11434/v1"
+export OPENAI_API_KEY="ollama"   # any non-empty value works; Ollama ignores it
+
+# 3. Run. Prefix the model with `openai/` so prefix routing selects the
+#    OpenAI-compatible provider, and use the exact Ollama tag.
+./target/release/claw --model openai/qwen3.5:9b
+```
+
+**The model name must match the Ollama tag exactly.** Ollama tags use a colon
+(`name:tag`), e.g. `qwen3.5:9b` — not a hyphen (`qwen3.5-9b`), which the server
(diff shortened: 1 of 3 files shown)


1716775457damn:
Great catch on the Ollama reasoning field semantics. The distinction between 
easoning, 
easoning_content, and 	hinking across providers is exactly the kind of API inconsistency that causes silent failures — good to see it handled in both streaming and non-streaming paths with tests.

One concern: other Ollama models (like DeepSeek-R1 distillations) may also emit content in the 
easoning field. The current fallback should handle those gracefully, but it might be worth adding a smoke test with deepseek-r1:7b or qwq:latest to confirm the extraction logic doesn't accidentally swallow content from models that use both 
easoning and content simultaneously. Some thinking models interleave reasoning chunks with content chunks in streaming mode.

code-yeongyu:
Superseded by #3236, which selectively landed the Ollama/Qwen reasoning-field behavior on top of #3234 without reintroducing the older routing/history changes.

Merged coverage in #3236 includes:
- non-streaming `reasoning` -> Thinking before text
- streaming `reasoning` -> Thinking before text
- existing `reasoning_content` regressions
- local/custom OpenAI-compatible routing regressions from #3234
- CLI Qwen tag hint and README pointer

#3236 CI is green and merged at 9b3548ca4337f54b001262474a35c209c17432b0.

code-yeongyu:
Closing as superseded by merged PR #3236.

1716775457damn:
Nice work on the Ollama integration. Surfacing the reasoning field for Qwen3 thinking models is critical - without it the output is just empty. The WSL setup guide in the README is a nice touch for Windows users. One thought: should we also mention the OLLAMA_HOST env var for users running Ollama on a different machine?

olabi:
Thanks, that's a great point! 🙏

You're right that anyone running Ollama on a different machine won't be reaching it at `127.0.0.1:11434` — in practice the host and port vary a lot (a LAN IP, a custom port, a reverse proxy, or WSL talking to the Windows host). Since there's no single "correct" URL to assume, documenting it in the README is exactly the right way to guide people.

Good news: `OLLAMA_HOST` already covers this. claw reads it and appends `/v1` (falling back to `http://127.0.0.1:11434` when unset), so pointing it at a remote box just works:

```bash
export OLLAMA_HOST="http://192.168.1.50:11434"
```

The superseding (now-merged) PR #3236 also adds an `OLLAMA_HOST` pointer to `rust/README.md` for this remote/custom-URL case, alongside the `OPENAI_BASE_URL` path — so it's covered on `main` today. Happy to expand that note with the explicit remote-host example above if you think it'd help readers.

1716775457damn:
Acknowledged — since this is superseded by #3236 which landed the reasoning-field behavior cleanly on top of #3234, closing the loop here. The multi-provider reasoning field handling (reasoning / reasoning_content / thinking) is solid groundwork regardless.

PR #3230, opened 2026-06-06
docs: add interactive session example to quick start

Gaurav-x111:
## Summary
- TBD

## Anti-slop triage
- Classification: 
- Evidence: 

## Verification
- [ ] No live secrets, tokens, private logs, or unrelated generated churn are included.

## Resolution gate

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ export ANTHROPIC_API_KEY="sk-ant-..."
 
 # 4. Run a prompt
 ./target/debug/claw prompt "say hello"
+
+# 5. Start an interactive session
+./target/debug/claw
 ```
 
 > [!NOTE]


1716775457damn:
Thanks for adding the interactive session example! A couple of suggestions to make it more useful for new users:

1. Consider adding a concrete multi-turn dialogue example (e.g., user asks a question → assistant follows up with a clarification → user confirms), so readers can see how the session state persists across turns.
2. It would help to note any prerequisites (e.g., auth setup, model selection) before starting the session, so users don't hit errors on first try.

The structure looks good overall — just needs a bit more detail to be truly actionable for first-time users.

1716775457damn:
Adding an interactive session example to the quick start is a good idea - it helps new users get hands-on faster. I noticed the PR summary is still marked as TBD; please fill in the description with what the example covers and why it was chosen. Also, consider linking to the relevant section in the main docs if the example mirrors an existing workflow.

1716775457damn:
The interactive session example is a good addition for onboarding. The suggestions about prerequisites (auth/model setup) would help prevent first-run confusion — worth adding before merge. Otherwise the structure is clean.

PR #3234, opened 2026-06-07
fix(providers): preserve OpenAI-compatible reasoning history

code-yeongyu:
## Summary
- repair OpenAI-compatible send/stream model handling so provider-prefixed routes send the stripped downstream model while internal events keep the original model
- persist non-streaming `Thinking` output blocks into CLI assistant events so reasoning history survives the next turn
- isolate two runtime tests that were failing from real global session state / invalid git metadata fixtures

## Validation
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p api deepseek_v4 -- --nocapture`
- `cargo test -p api`
- `cargo test -p runtime`
- `cargo test -p rusty-claude-cli response_to_events_renders_collapsed_thinking_summary -- --nocapture`
- `cargo clippy -p api --all-targets` (passes with pre-existing warnings)
- `cargo test --workspace`

## Manual QA
- `cargo run -p rusty-claude-cli -- --help` rendered the CLI help surface successfully

## Notes
- The first commit is the provider/CLI behavior fix.
- The second commit is test isolation only, keeping the behavior fix independently reviewable.

### DIFF
--- a/rust/crates/api/src/providers/mod.rs
+++ b/rust/crates/api/src/providers/mod.rs
@@ -296,9 +296,6 @@ pub fn metadata_for_model(model: &str) -> Option<ProviderMetadata> {
     None
 }
 
-
-
-
 #[must_use]
 pub fn strip_provider_prefix(canonical_model: &str) -> String {
     if let Some(pos) = canonical_model.find('/') {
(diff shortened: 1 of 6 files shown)


1716775457damn:
Preserving reasoning history for OpenAI-compatible providers is important for multi-turn conversations. This + #3236 should fix the build break from #3235. Looks solid.

PR #3237, opened 2026-06-08
fix: validate attached redirection paths

hiSandog:
## Summary
- strip attached shell redirection operators before extracting path candidates
- validate redirection targets like <../file and 2>../file against workspace scope
- add regression coverage for attached redirection targets

## Validation
- python3 -m unittest tests.test_security_scope -q
- python3 -m compileall -q src/path_scope.py tests/test_security_scope.py
- git diff --check -- src/path_scope.py tests/test_security_scope.py


1716775457damn:
Good catch on the redirection bypass. Stripping shell redirects before path extraction is the right fix. One minor suggestion: confirm this handles no-space-after-redirect like 2>../file (valid in bash). Otherwise LGTM.

1716775457damn:
Good catch on the attached redirection operators — <../file and 2>../file are real edge cases that could slip through naive path validation. The regression tests look solid. One suggestion: consider also covering >> (append) and &> (combined stdout+stderr) in the stripping logic, since those are also valid shell redirection syntax that could be abused similarly.

1716775457damn:
Merged — the >> and &> coverage suggestion can be a follow-up since the core fix (stripping redirects before path extraction) is already in. Good to see this landed quickly.

PR #3239, opened 2026-06-08
fix(commands): wire workspace slash command

AbbyJL:
## Summary
- Wired `/workspace` into the slash-command parser and CLI handling.
- Added `/cwd` as a supported alias.
- Implemented workspace reporting and optional in-root directory switching for resumed/live CLI sessions.
- Added parser coverage for `/workspace` and `/cwd`.

## Breaking changes
- None expected. The change is additive.
- `workspace`/`cwd` can only switch into paths under the current workspace root; attempts to leave the root return an explicit error and suggest `claw --cwd <path>`.

## Verification
- `cargo test -p commands --lib`
- `rustfmt` on the touched Rust files
- `cargo check -p rusty-claude-cli --tests` was blocked by an unrelated pre-existing syntax error in `crates/api/src/providers/openai_compat.rs`

### DIFF
--- a/rust/crates/commands/src/lib.rs
+++ b/rust/crates/commands/src/lib.rs
@@ -1131,6 +1131,9 @@ pub enum SlashCommand {
     SecurityReview,
     Keybindings,
     PrivacySettings,
+    Workspace {
+        path: Option<String>,
+    },
     Plan {
         mode: Option<String>,
     },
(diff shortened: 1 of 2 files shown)


1716775457damn:
Good fix — wiring the workspace slash command is needed for discoverability. Noticed there are merge conflicts that need resolving before review can proceed. Also, the PR summary is still TBD; please fill in what commands are wired, any breaking changes, and steps to verify.

PR #3240, opened 2026-06-08
Add route.bible skill

dpshde:
This PR adds the skill definition for route.bible, enabling the agent to handle app-agnostic Bible passage routing.

### DIFF
--- a/skills/route.bible/skill.md
+++ b/skills/route.bible/skill.md
@@ -1,0 +1,14 @@
+# route.bible/skill.md
+
+## Overview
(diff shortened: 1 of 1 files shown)


1716775457damn:
Nice addition. The URL contract using standard book abbreviations and OSIS-style references is clean and consistent with the existing skill conventions.

One thought: for offline or rate-limited scenarios, would it make sense to mention in the description how the agent should fall back (e.g., suggesting the user open the link manually or using a local lookup)?

The ?qr=true integration is a nice touch for mobile sharing.

PR #3243, opened 2026-06-11
fix: prevent path traversal via session_id in session_store

mlzoo:
## What
`load_session` / `save_session` in `src/session_store.py` built the session file path directly from the caller-supplied `session_id` with no validation.

## Impact
A `session_id` such as `../../../../tmp/secrets` — reachable from the `load-session` CLI subcommand (`src/main.py:168`) — let a caller read JSON files outside the `.port_sessions` store, and `save_session` could likewise write outside it. (Path traversal is listed as in-scope in `SECURITY.md`.)

## Fix
Add `_validate_session_id()` which rejects anything outside `[A-Za-z0-9_-]` (path separators, `..`, absolute and Windows paths) before the id is used to build a filename. Both `save_session` and `load_session` validate first. Legitimate ids (uuid4 hex, `session-<digits>-<n>`) are unaffected.

## Tests
Adds `SessionStorePathTraversalTests` to `tests/test_security_scope.py`:
- traversal id rejected on load
- absolute/separator/`..` ids rejected
- traversal id rejected on save
- legitimate id round-trips (no regression)

Full suite: `python3 -m unittest discover -s tests` -> 50 passed.

### DIFF
--- a/src/session_store.py
+++ b/src/session_store.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import re
 from dataclasses import asdict, dataclass
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


1716775457damn:
Nice catch on the path traversal. The whitelist approach ([A-Za-z0-9_-]) is definitely the right call over trying to sanitize dots and slashes — much simpler and less error-prone.

One minor consideration: the current allowlist excludes common session-id characters like hyphens in non-leading position. Looking at the test description it seems hyphens are already covered (_ is included), but just confirming — are there any existing session IDs in the wild that use additional safe characters like dots in a flat namespace (e.g., user.session.1)? If not, this is clean.

Security impact is clear and the test coverage looks solid. Good fix.

mlzoo:
Thanks! To confirm — no existing session IDs use dots. The two formats in the codebase are uuid4 hex strings (e.g. `a1b2c3d4e5f6...`) and `session-<digits>-<n>`, both of which are fully covered by `[A-Za-z0-9_-]`. So no legitimate IDs are rejected by this allowlist.

1716775457damn:
Confirmed — the whitelist approach is the right call for path traversal prevention. Since all existing session ID formats (uuid4 hex and session-<digits>-<n>) are covered by [A-Za-z0-9_-], there should be no regression risk. The test coverage for traversal rejection plus legitimate round-trip is thorough. LGTM on the fix.

mlzoo:
Thanks for confirming — keeping the whitelist approach as-is.

1716775457damn:
The fix looks complete — validation covers all traversal vectors (.., absolute paths, separators) and the legit session ID formats are preserved. The test coverage for both rejection and round-trip cases is thorough. Ready to merge from my side.

PR #3245, opened 2026-06-12
Allow null device redirection in path scope

hiSandog:
## Summary
- allow WorkspacePathScope to treat /dev/null as a null-device sink instead of an out-of-workspace path
- add coverage for echo ok >/dev/null

## Validation
- python3 -m unittest tests.test_security_scope
- python3 -m unittest discover tests

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -42,6 +42,13 @@ Out of scope:
 - reports against third-party providers or upstream tools without a Claw Code
   integration issue.
 
+## Workspace path scope
+
+Claw Code validates shell and file-tool path operands against the active
+workspace roots. Redirection to `/dev/null` is treated as a null-device sink,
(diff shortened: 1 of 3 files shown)


1716775457damn:
LGTM on the approach. Treating /dev/null as a null-device sink in path scope makes sense since it's a side-effect-free operation. One suggestion: consider documenting this behavior clearly in the security scope docs so users understand why /dev/null is exempt from path restrictions.

hiSandog:
Documented this in 70f8e6d by adding a Workspace path scope note to SECURITY.md. It now explains that /dev/null is allowed because it is treated as a null-device sink rather than a filesystem target outside the workspace.

hiSandog:
Closing this as stale under my PR maintenance rule: it has been open for more than 15 full days. I am leaving the branch untouched.

PR #3250, opened 2026-06-17
feat(team): team enhancements, subagentModel, parallel tool exec

TheArchitectit:
## Summary

Ports the agent-team enhancement layer onto main so the model can spin up
coordinated sub-agent teams for parallel work. **2,095 lines added across 11 files
— zero deletions of existing upstream functionality.**

Single clean commit built from `upstream/main`. No TUI, LSP, or setup-wizard
contamination from other in-flight work.

---

## Motivation

The `TeamCreate`, `TeamDelete`, and `Agent` tools already existed, but they
didn't work end-to-end because:

1. **`subagentModel` was silently dropped** — `/setup` saved the sub-agent model
   into `settings.json` but `RuntimeConfig` never stored or exposed it, so
   agents always fell back to a hard-coded default.
2. **Sub-agents couldn't reach the user's provider** — the main CLI calls
   `inject_config_as_env_fallbacks()` to set `CLAWCUSTOMOPENAI_API_KEY` /
   `CLAWCUSTOMOPENAI_BASE_URL` from `/setup`-saved config, but the sub-agent
   spawn path in `tools` never did, so custom providers were invisible.
3. **`providerFallbacks.primary` silently overrode the agent's model** — the
   configured fallback primary replaced whatever model `resolve_agent_model`
   picked, routing agents to a dead model instead of the session's provider.
4. **404 model-not-found killed the chain** — a dead primary was not retryable,
   so the fallback chain never advanced to working models.

This PR fixes all four and adds the team coordination tools the model needs
to orchestrate multi-agent work.

---

### 1. `subagentModel` config wiring (bug fix)

The `subagentModel` field was validated by `config_validate.rs` but **never
stored or read** — the setting from `/setup` was silently dropped.

| Change | File |
|--------|------|
| `RuntimeFeatureConfig.subagent_model` field | `runtime/src/config.rs` |
| `RuntimeConfig::subagent_model()` accessor | `runtime/src/config.rs` |
| `parse_optional_subagent_model()` (reads `subagentModel` / `subagent_model`, blank → `None`) | `runtime/src/config.rs` |
| `runtime::inject_config_as_env_fallbacks()` — moved from private `main.rs` fn to `pub fn` in runtime so sub-agents can call it | `runtime/src/config.rs` + `runtime/src/lib.rs` |
| `resolve_agent_model` fallback chain: explicit model → `subagentModel` → session provider model → `DEFAULT_AGENT_MODEL` | `tools/src/lib.rs` |

### 2. Parallel tool execution (perf)

When the model emits multiple `tool_use` blocks in one response, read-only
tools now run **concurrently** via `std::thread::scope` instead of
sequentially.

**Parallel-safe tools:** `read_file`, `glob_search`, `grep_search`,
`WebFetch`, `WebSearch`, `ToolSearch`, `Skill`, `LSP`, `GitStatus`,
`GitDiff`, `GitLog`, `GitShow`, `Agent`, `AgentMessage`, `AgentSuggestion`,
`TeamStatus`, `TeamInfo`, `TaskGet`, `TaskList`, `TaskOutput`

**Sequential tools:** everything else (writes, side-effects)

| Change | File |
|--------|------|
| `ToolCall`, `ToolResult`, `TurnProgressReporter` re-exported from runtime | `runtime/src/conversation.rs` + `runtime/src/lib.rs` |
| `ToolExecutor::execute_batch` default impl on trait | `runtime/src/conversation.rs` |
| `CliToolExecutor::execute_batch` override with parallel-safe classification | `rusty-claude-cli/src/main.rs` |

### 3. Team coordination layer (feature)

New tools and capabilities for multi-agent work:

| Tool | Purpose |
|------|---------|
| `TeamCreate` | Spawn N agents (1–6× presets: tiny/small/medium/large/xlarge/mega) with a shared prompt |
| `TeamDelete` | Mark a team deleted, clean inbox, release task claims |
| `TeamStatus` | Live snapshot, final summary, or event inbox for a team |
| `AgentMessage` | Send/receive messages between agents (shared mailbox) |
| `TaskClaim` | Claim/list/complete tasks to prevent duplicate work across agents |
| `AgentSuggestion` | Agents suggest AGENTS.md patterns/pitfalls |

Supporting infrastructure:

- **Shared mailbox directory** at `~/.clawd-agents/mailbox/team/{team_id}/`
- **Background team watcher thread** — polls agent `.json` files every 2s,
  prints `[team]` progress to stderr, writes events to `-events.jsonl`
- **Mode presets** — `tiny`/`1x` through `mega`/`6x` controlling agent count
  and role distribution (Explore, Plan, Verification, Reviewer)
- **`/team` slash command** — `on|off|status` to toggle `CLAWD_AGENT_TEAMS`
  (required by `TeamCreate`)
- **Enriched tool descriptions** on `TeamCreate` and `Agent` steering the
  model toward correct usage

### 4. Model-resolution robustness fixes (bug fix)

| Fix | Detail |
|-----|--------|
| **`qualify_for_provider()`** | Bare model names (e.g. `claude-haiku-4-5-20251001`) get `custom/` prefix when the active provider is `custom-openai`, so sub-agents route through the same endpoint as the main session |
| **`providerFallbacks.primary` precedence** | The caller's resolved model is now the **primary**; `providerFallbacks.primary` + `fallbacks` are appended as recovery entries, deduped. Previously the config primary silently replaced the caller's model |
| **404 model-not-found chain fallthrough** | Added `fallback_chain_eligible()` that treats 404/400 "not found" as chain-eligible, so a dead configured primary advances to `kimi`/`qwen`/etc. Added `ApiError::status_code()` and `response_body()` accessors for the detection |
| **Sub-agents inject `/setup` credentials** | `run_agent_job()` calls `runtime::inject_config_as_env_fallbacks()` before building the agent runtime, so custom providers set up via `/setup` work for spawned agents |

### 5. Also fixed

| Fix | File |
|-----|------|
| `lane_completion.rs` test helper set stale `team_id`/`task_id` fields | `tools/src/lane_completion.rs` |
| `/team` slash command had no `validate_slash_command_input` match arm (parsed as `Unknown`) | `commands/src/lib.rs` |
| `ApiError` gains `status_code()` and `response_body()` public accessors | `api/src/error.rs` |
| `model_token_limit` catch-all: `_ => Some(...)` → `_ => None` — unknown models skip preflight, let API enforce limits (caught by @Offwhite-Del) | `api/src/providers/mod.rs` |
| `Lsp` variant added to match arms | `commands/src/lib.rs` + `rusty-claude-cli/src/main.rs` |
| `DEFAULT_CUSTOM_OPENAI_BASE_URL` constant added | `api/src/providers/openai_compat.rs` |
| Provider chain precedence test updated to match new behavior | `tools/src/lib.rs` |

---

### Build
```
cargo check --workspace              ✅
cargo build -p rusty-claude-cli --release  ✅
```

### Tests
```
tools/src/lib.rs           112 passed  (incl. 3 fallback_chain_tests, 4 provider_runtime_client chain tests)
commands/src/lib.rs         42 passed  (incl. /team parse regression)
runtime team_cron            17 passed
rusty-claude-cli bin        207 passed
```

### Live testing (IdleDefense repo, Godot 4.6 game)
- `/team on` → TeamCreate with 4 agents → all agents spawned, claimed tasks, read files
- Agents used `custom/openclaw` provider (via `qualify_for_provider`)
- `TeamStatus` showed live progress, inbox messages, task claims
- 2/4 agents completed full reports (Explore + Verification), 2/4 hit iteration limits on complex analysis prompts (expected — increase `max_iterations` or simplify prompts for large repos)

---

## Diff stats

```
 rust/crates/api/src/error.rs             |   43 +
 rust/crates/api/src/lib.rs               |    3 +-
 rust/crates/api/src/providers/mod.rs     |   39 +-
 rust/crates/api/src/providers/openai_compat.rs | 1 +
 rust/crates/commands/src/lib.rs          |   28 +-
 rust/crates/runtime/src/config.rs        |   70 +
 rust/crates/runtime/src/conversation.rs  |  203 +-
 rust/crates/runtime/src/lib.rs           |    5 +-
 rust/crates/rusty-claude-cli/src/main.rs |    3 +
 rust/crates/tools/src/lane_completion.rs |    2 +
 rust/crates/tools/src/lib.rs             | 1699 ++++++++++++++++++++++++
 11 files changed, 2095 insertions(+), 113 deletions(-)
```

## Key files

| File | Lines | What it does |
|------|-------|-------------|
| `tools/src/lib.rs` | +1,699 | All team tool logic: TeamCreate, TeamDelete, TeamStatus, AgentMessage, TaskClaim, AgentSuggestion, mode presets, mailbox, watcher, qualify_for_provider, fallback_chain_eligible |
| `runtime/src/conversation.rs` | +203 | ToolCall/ToolResult types, TurnProgressReporter trait, execute_batch default impl |
| `runtime/src/config.rs` | +70 | subagent_model field/accessor/parser, inject_config_as_env_fallbacks |
| `api/src/error.rs` | +43 | status_code()/response_body() accessors for chain-fallback detection |
| `api/src/providers/mod.rs` | +39 | model_token_limit generic fallback, custom-openai provider entry |

---

## What this does NOT include

This PR is **purely additive** on top of upstream/main. It does **not** include:
- TUI / split-pane dashboard changes (separate PR)
- LSP module restructuring (already merged)
- Setup wizard / provider config changes (already merged)
- Custom-openai namespace routing (separate PR)
- Hooks refactoring (separate PR)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ archive/
 status-help.txt
 # Legacy Python port session scratch artifacts
 .port_sessions/
+.claude/
--- a/USAGE.md
+++ b/USAGE.md
@@ -364,10 +364,11 @@ Reasoning variants (`qwen-qwq-*`, `qwq-*`, `*-thinking`) automatically strip `te
 | **xAI** | OpenAI-compatible | `XAI_API_KEY` | `XAI_BASE_URL` | `https://api.x.ai/v1` |
 | **OpenAI-compatible** | OpenAI Chat Completions | `OPENAI_API_KEY` | `OPENAI_BASE_URL` | `https://api.openai.com/v1` |
 | **DashScope** (Alibaba) | OpenAI-compatible | `DASHSCOPE_API_KEY` | `DASHSCOPE_BASE_URL` | `https://dashscope.aliyuncs.com/compatible-mode/v1` |
+| **Custom OpenAI-compat** | OpenAI Chat Completions | `CLAWCUSTOMOPENAI_API_KEY` | `CLAWCUSTOMOPENAI_BASE_URL` | none (must be supplied) |
 
 The OpenAI-compatible backend also serves as the gateway for **OpenRouter**, **Ollama**, and any other service that speaks the OpenAI `/v1/chat/completions` wire format — just point `OPENAI_BASE_URL` at the service.
 
-**Model-name prefix routing:** If a model name starts with `openai/`, `local/`, `gpt-`, `qwen/`, `qwen-`, `kimi/`, or `kimi-`, the provider is selected by the prefix regardless of which env vars are set. This prevents accidental misrouting to Anthropic when multiple credentials exist in the environment. For the default OpenAI API and local/private OpenAI-compatible endpoints, `openai/` is a routing prefix and is stripped before the request hits the wire. For non-local custom `OPENAI_BASE_URL` gateways, slash-containing OpenAI-compatible slugs (for example OpenRouter-style `openai/gpt-4.1-mini`) are preserved so the gateway receives the model ID it expects. The `local/` prefix is an explicit escape hatch for local slash-containing model IDs: it is stripped while the rest of the model ID is sent verbatim.
+**Model-name prefix routing:** If a model name starts with `openai/`, `local/`, `custom/`, `gpt-`, `qwen/`, `qwen-`, `kimi/`, or `kimi-`, the provider is selected by the prefix regardless of which env vars are set. This prevents accidental misrouting to Anthropic when multiple credentials exist in the environment. For the default OpenAI API and local/private OpenAI-compatible endpoints, `openai/` is a routing prefix and is stripped before the request hits the wire. For non-local custom `OPENAI_BASE_URL` gateways, slash-containing OpenAI-compatible slugs (for example OpenRouter-style `openai/gpt-4.1-mini`) are preserved so the gateway receives the model ID it expects. The `local/` prefix is an explicit escape hatch for local slash-containing model IDs: it is stripped while the rest of the model ID is sent verbatim. The `custom/` prefix selects the Claw-specific custom OpenAI-compat provider (`CLAWCUSTOMOPENAI_*` env vars) and is also stripped on the wire; use it when you want a private OpenAI-compatible endpoint that does not collide with real `OPENAI_*` credentials.
 
 ### Tested models and aliases
 
@@ -407,15 +408,22 @@ Local project settings override user-level settings. Aliases resolve through the
 
 Model selection precedence is CLI flag, environment, config, then default. The environment model slot accepts `CLAW_MODEL`, `ANTHROPIC_MODEL`, and `ANTHROPIC_DEFAULT_MODEL` in that order; aliases from those variables are resolved and validated before provider startup. `claw --output-format json status` exposes `model_raw`, `model_alias_resolved_to`, and `model_env_var` so automation can see the winning value.
 
+### /setup wizard
+
+Run `/setup` inside the REPL (or `claw setup` from the shell) to save provider credentials interactively to `~/.claw/settings.json`. Saved settings are injected as env-var fallbacks at startup, preserving the precedence: env var > `.env` file > stored config.
+
+For a **Custom (OpenAI-compat)** endpoint, `/setup` now stores `kind: "custom-openai"` and sets the dedicated `CLAWCUSTOMOPENAI_API_KEY` / `CLAWCUSTOMOPENAI_BASE_URL` env vars. This lets a private OpenAI-compatible proxy coexist with real OpenAI, NeuralWatt, or other tools that already consume `OPENAI_*`. A bare model name saved by `/setup` is normalized to `custom/<model>` and stripped back to the bare id on the wire.
+
 ### How provider detection works
 
 1. If the resolved model name starts with `claude` → Anthropic.
 2. If it starts with `grok` → xAI.
-3. If it starts with `openai/`, `local/`, or `gpt-` → OpenAI-compatible.
+3. If it starts with `openai/`, `local/`, `custom/`, or `gpt-` → OpenAI-compatible.
 4. If it starts with `qwen/`, `qwen-`, `kimi/`, or `kimi-` → DashScope-compatible OpenAI wire format.
 5. If `OPENAI_BASE_URL` is set, local-looking unknown model names such as `llama3.2` or `qwen2.5-coder:7b` route to the OpenAI-compatible client for local/gateway servers.
-6. Otherwise, `claw` checks which credential is set: Anthropic first, then OpenAI, then xAI. If only `OPENAI_BASE_URL` is set, it still routes to OpenAI-compatible for authless local servers.
-7. If nothing matches, it defaults to Anthropic.
+6. If it starts with `custom/` → the Claw custom OpenAI-compat provider (`CLAWCUSTOMOPENAI_API_KEY` / `CLAWCUSTOMOPENAI_BASE_URL`).
+7. Otherwise, `claw` checks which credential is set: Anthropic first, then OpenAI, then xAI. If only `OPENAI_BASE_URL` is set, it still routes to OpenAI-compatible for authless local servers.
+8. If nothing matches, it defaults to Anthropic.
 
 
 ### Provider diagnostics and custom OpenAI-compatible parameters
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -30,12 +30,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "allocator-api2"
-version = "0.2.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
-
-[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -362,25 +356,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
-name = "cassowary"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df8670b8c7b9dae1793364eafadf7239c40d669904660c5960d74cfd80b46a53"
-
-[[package]]
 name = "cast"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
-
-[[package]]
-name = "castaway"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
-dependencies = [
- "rustversion",
-]
 
 [[package]]
 name = "cc"
@@ -543,20 +522,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "compact_str"
-version = "0.8.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3b79c4069c6cad78e2e0cdfcbd26275770669fb39fd308a752dc110e83b9af32"
-dependencies = [
- "castaway",
- "cfg-if",
- "itoa",
- "rustversion",
- "ryu",
- "static_assertions",
-]
-
-[[package]]
 name = "compat-harness"
 version = "0.1.3"
 dependencies = [
@@ -651,15 +616,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-channel"
-version = "0.5.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
-dependencies = [
- "crossbeam-utils",
-]
-
-[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -731,18 +687,8 @@ version = "0.20.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc7f46116c46ff9ab3eb1597a45688b6715c6e628b5c133e288e709a29bcb4ee"
 dependencies = [
- "darling_core 0.20.11",
- "darling_macro 0.20.11",
-]
-
-[[package]]
-name = "darling"
-version = "0.23.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
-dependencies = [
- "darling_core 0.23.0",
- "darling_macro 0.23.0",
+ "darling_core",
+ "darling_macro",
 ]
 
 [[package]]
@@ -760,36 +706,12 @@ dependencies = [
 ]
 
 [[package]]
-name = "darling_core"
-version = "0.23.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9865a50f7c335f53564bb694ef660825eb8610e0a53d3e11bf1b0d3df31e03b0"
-dependencies = [
- "ident_case",
- "proc-macro2",
- "quote",
- "strsim",
- "syn",
-]
-
-[[package]]
 name = "darling_macro"
 version = "0.20.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc34b93ccb385b40dc71c6fceac4b2ad23662c7eeb248cf10d529b7e055b6ead"
 dependencies = [
- "darling_core 0.20.11",
- "quote",
- "syn",
-]
-
-[[package]]
-name = "darling_macro"
-version = "0.23.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac3984ec7bd6cfa798e62b4a642426a5be0e68f9401cfc2a01e3fa9ea2fcdb8d"
-dependencies = [
- "darling_core 0.23.0",
+ "darling_core",
  "quote",
  "syn",
 ]
@@ -818,7 +740,7 @@ version = "0.20.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2d5bcf7b024d6835cfb3d473887cd966994907effbe9227e8c8219824d06c4e8"
 dependencies = [
- "darling 0.20.11",
+ "darling",
  "proc-macro2",
  "quote",
  "syn",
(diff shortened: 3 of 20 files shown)


Offwhite-Del:
**发现并修复: preflight_skips_unknown_models 测试失败**

**根因:** `model_token_limit()` catch-all arm (`_ =>`) 为未知模型分配了默认 context window (131072)，导致 `preflight_message_request` 对未知模型返回 `ContextWindowExceeded`。

**修复:** catch-all 应返回 `None`，`preflight_message_request` 已正确处理 `None` → 跳过预检。

修改: `crates/api/src/providers/mod.rs`, change `_ => Some(ModelTokenLimit {...})` to `_ => None` (line ~694)

**验证:** ✅ `cargo test preflight` — 4/4 passed
**修复分支:** `Offwhite-Del/sun:fix/preflight-unknown-models`

建议直接 rebase 到最新 main (main 上已经是 `_ => None`)，或者 Cherry-pick 这个 commit: `6e28598`

TheArchitectit:
感谢发现这个问题！你说得对，catch-all 应该返回 `None` 让 preflight 跳过未知模型。已经修复并推送上来了。谢谢！🙏

1716775457damn:
Nice work on the subagentModel fix — that was a real gap where /setup saved the model but RuntimeConfig never surfaced it. The team coordination layer looks clean. One thing: cargo fmt is failing, you'll want to run cargo fmt --all and push before this can land.

1716775457damn:
Ambitious feature set — subagentModel and parallel tool execution would be great additions. Noting that 2 CI checks are failing (build + test), likely need a rebase on recent main changes. Thanks @TheArchitectit!

TheArchitectit:
## ⚠️ Dependency + rebase notes (force-pushed)

This PR was rebased `--onto` current `upstream/main`. The branch now contains **4 commits**:

1. `fefa608c` fix: inject /setup provider settings as env var fallbacks for API client
2. `2d6540c5` fix(setup): apply saved provider url/key/model for custom OpenAI endpoints
3. `0c339e7b` feat(setup): separate env namespace for custom OpenAI-compat provider
4. `224aff84` feat(team): team enhancements, subagentModel, parallel tool exec  ← **the actual team changes**

### Why commits 1–3 are here

Commits 1–3 are **identical** to PR #3248 (`feat/custom-openai-namespace-upstream`) — the custom-OpenAI provider namespace work. They are included here **only as a base** so this PR builds clean and reviewable today. The team layer in commit 4 depends on `runtime::inject_config_as_env_fallbacks()` (moved from a private `main.rs` fn into `runtime/src/config.rs` as a `pub fn` so sub-agent threads can reach it).

### Merge order (important)

**#3248 must merge first.** After #3248 lands on `upstream/main`, this branch should be re-rebased to drop commits 1–3 (they will already be on main) so only commit 4 remains. I do not have merge rights, so the rebase drop will be a follow-up force-push once #3248 is in.

```
git rebase --onto upstream/main 0c339e7b feat/team-pr-clean
```

### What commit 4 (the team work) contains

Only team-specific changes — no custom-OpenAI provider implementation is duplicated:
- `subagentModel` config field wiring (`RuntimeFeatureConfig.subagent_model`, `RuntimeConfig::subagent_model()`)
- `runtime::inject_config_as_env_fallbacks()` `pub fn` + re-export
- `ToolExecutor::execute_batch` + `ToolCall`/`ToolResult`/`TurnProgressReporter` (parallel read-only tool execution via `std::thread::scope`)
- `TeamCreate` mode presets, `TeamStatus`, `AgentMessage`, `TaskClaim`, `AgentSuggestion`, `ContextRequest` tools
- `/team [on|off|status]` slash command, Ctrl+T toggle, ProviderSwap handler
- `run_turn_to<W: Write>` refactor
- `ApiError::status_code()` / `response_body()` accessors
- `model_token_limit` additions: `qwen3.*`, `glm-5.1-fast`, generic 128K/8K fallback

### Build status

`cargo check --workspace` passes on top of `upstream/main` + the 3 base commits. Pre-existing clippy warnings on `upstream/main` are unchanged by this PR.

1716775457damn:
Thanks for the detailed rebase notes and the dependency clarification. The merge order plan (merge #3248 first, then rebase this to drop commits 1-3) makes sense and is the right approach to keep history clean.

One concern: cargo fmt is still failing, which is a trivial fix — could you run cargo fmt --all and push before this lands? Also, the build and test failures may need investigation beyond just a rebase; would be good to confirm they pass after the fmt fix and re-rebase.

The subagentModel wiring through RuntimeFeatureConfig → RuntimeConfig looks solid, and the custom-openai namespace separation with dedicated env vars (CLAWCUSTOMOPENAI_*) is a clean design choice. The parallel tool execution via std::thread::scope for read-only tools is a pragmatic approach — looking forward to seeing this land.

TheArchitectit:
Thanks for the review! All addressed:

- **`cargo fmt --all`** — ran and pushed (was just 2 trailing blank lines in `main.rs`)
- **`model_token_limit` catch-all** — the fix from @Offwhite-Del was lost during rebase; restored it (`_ => None` so unknown models skip preflight)
- **Build** — `cargo check --workspace` passes (8 dead-code warnings, no errors)
- **Tests** — all passing (160 API tests, 112 tools tests, 42 commands tests, 207 CLI tests)

Ready for #3248 to merge so I can rebase and drop commits 1-3 as discussed.

1716775457damn:
The architecture decision to move inject_config_as_env_fallbacks from main.rs into the runtime crate is the right call — sub-agents need provider visibility at spawn time, not just at CLI entry. One thing I'd suggest: the subagentModel wiring through RuntimeFeatureConfig looks clean, but consider adding a unit test that explicitly verifies the subagent_model -> subagentModel() accessor chain end-to-end with both snake_case and camelCase config keys, since that was the original silent-drop bug. Otherwise LGTM on the design.

TheArchitectit:
Good suggestion — added 3 unit tests for the `subagentModel` accessor chain:

- `subagent_model_reads_camel_case_config_key` — verifies `"subagentModel"` camelCase key flows end-to-end to `config.subagent_model()`
- `subagent_model_reads_snake_case_config_key` — verifies `"subagent_model"` snake_case alias works too
- `subagent_model_blank_value_returns_none` — verifies blank strings return `None` (falls back to default model)

All 3 pass. Pushed in commit 30ff4edf.

1716775457damn:
The 3 unit tests for subagentModel accessor look solid — covering camelCase, snake_case, and blank-value fallback addresses the original silent-drop bug precisely. Moving inject_config_as_env_fallbacks into the runtime crate was the right architectural call; sub-agents need provider visibility at spawn time, not just CLI entry. Once #3248 merges and the rebase drops commits 1-3, this should be ready to land.

1716775457damn:
Good work on this PR. The subagentModel wiring fix closes a real gap — having /setup save the model config but RuntimeConfig silently drop it was a frustrating bug for users trying to configure sub-agents. Moving inject_config_as_env_fallbacks into the runtime crate is architecturally sound; sub-agents need provider visibility at spawn time. The 3 unit tests covering camelCase, snake_case, and blank-value for the accessor chain are thorough and directly address the original silent-drop issue. Once #3248 lands and this gets rebased to drop commits 1-3, this should be ready to merge.

TheArchitectit:
## Review fixes — pushed to `feat/team-pr-clean`

Commit `6ee9e03` addresses three issues found during review:

### 1. ✅ Workspace `unsafe_code` lint restored to `"forbid"`

The PR had downgraded the workspace-level lint from `forbid` to `warn` to support the `gag` crate's `dup`/`dup2` usage in the TUI. This weakened safety guarantees across **all** crates unnecessarily.

**Fix:** Restored `unsafe_code = "forbid"` at the workspace level in `rust/Cargo.toml`. The per-crate `[lints.rust] unsafe_code = "allow"` on `rusty-claude-cli` already overrides it where needed — that override is unchanged.

@Offwhite-Del — this preserves the safety boundary you flagged during earlier review. No other crate is affected.

### 2. ✅ TaskClaim TOCTOU race eliminated

`claim_task()` had a classic time-of-check-to-time-of-use race:

```rust
// BEFORE (racy)
if lock_path.exists() {
    return Ok(false);
}
// ... write tmp file, then rename
```

Two agents could both pass `exists()`, both write temp files, and both attempt `rename()`. On Linux, `rename` is atomic but **overwrites** — so the second agent's rename would silently overwrite the first agent's claim.

**Fix:** Replaced with `OpenOptions::new().create_new(true).open(&lock_path)` — the `create_new` flag makes the open itself atomic. If the file already exists, it returns `AlreadyExists` immediately. No race window.

```rust
// AFTER (atomic)
match std::fs::OpenOptions::new()
    .write(true)
    .create_new(true)
    .open(&lock_path)
{
    Ok(mut file) => {
        file.write_all(content.as_bytes())?;
        Ok(true)
    }
    Err(ref e) if e.kind() == AlreadyExists => Ok(false),
    Err(e) => Err(e.to_string()),
}
```

@1716775457damn — this was the most concerning issue from the review. The temp-file-then-rename approach looked atomic but wasn't on Linux.

### 3. ✅ TUI scaffolding removed (future PR)

The PR description says it doesn't include TUI changes, but several TUI dependencies and a `/tui` slash command had leaked in:

**Removed from `rusty-claude-cli/Cargo.toml`:**
- `ratatui = "0.29"`
- `tui-textarea = "0.7"`
- `gag = "1"`
- `crossbeam-channel = "0.5"`
- `libc = "0.2"`
- `unicode-width = "0.2"`

**Removed from `commands/src/lib.rs`:**
- `/tui` slash command spec

None of these were referenced anywhere in the source code — they were scaffolded for a future TUI PR. Cargo.lock regenerated (248 transitive dep entries cleaned out). Also updated the `slash_command_specs().len()` test assertion from 141 → 140.

### Also: provider chain test updated

The `provider_runtime_client_chain_primary_override_replaces_constructor_model` test was updated to reflect the new fallback chain behavior where the caller's resolved model is prepended as the chain primary (chain length 2 → 3, with `claude-haiku-4-5-20251213` as `chain[0]`).

### Verification

```
cargo check --workspace     ✅
cargo fmt --check           ✅
cargo test -p tools         114 passed, 1 failed*
cargo test -p commands      42 passed
cargo test -p runtime       17 passed
```

\* `repl_executes_python_code` — pre-existing failure, missing python runtime in sandbox, unrelated to these changes.

1716775457damn:
Thanks for addressing the review feedback — restoring the workspace-level forbid lint and deduplicating the config logic looks clean. The fallback-to-injected-env pattern in subagent initialization is a solid safety net. :+1:

1716775457damn:
Team enhancements with subagentModel and parallel tool execution is a solid feature addition. The ability to configure per-team subagent models will help with cost optimization and specialization.

1716775457damn:
Two details worth pinning down for the parallel read-only execution: (1) result ordering — if tool results are appended in completion order rather than call order, the transcript becomes non-deterministic across runs for the same input, which hurts reproducibility and makes failures hard to replay; keying results to the original call index would keep it stable. (2) a concurrency bound — unbounded fan-out against a local model server will queue or thrash, so a configurable cap with a sane default (and a per-team override alongside subagentModel) would be safer. Also worth confirming the workspace-scope check runs per tool call rather than once for the whole batch, since parallelism makes it easy to smuggle one out-of-scope path among several valid ones.

TheArchitectit:
All three points addressed in 6ee9e03..697911b (`feat/team-pr-clean`):

**1. Deterministic ordering — was already index-keyed, now airtight + tested**
`execute_batch` sorts results into original call slots and never appends in completion order. The real hole was panic handling: a panicking worker thread used to fall back to index 0 (or leave a gap). Each spawn now carries its own fallback tuple `(idx, tool_use_id, tool_name)`, so even a panic lands in its correct slot. Regression test: `parallel_batch_results_follow_call_index_even_with_workspace_escapes` runs 6 reads at cap=2 (forcing two waves with interleaved completion) and asserts `results[i].tool_use_id == t{i}` and correct per-slot content.

Documented parity caveat in-code: parallel-safe calls execute before sequential ones within a batch (slots still call-order), so models must not emit intra-batch dependent calls.

**2. Bounded fan-out — cap + per-team override**
- `maxParallelToolCalls` (also `max_parallel_tool_calls`) in settings; `CLAWD_PARALLEL_TOOL_CALLS` env; default **8**. Parallel-safe calls run in bounded waves via `thread::scope`.
- Per-team: `TeamCreate.maxConcurrentAgents` input, falling back to the `maxConcurrentAgents` setting, falling back to unlimited (pre-fix behavior preserved). Enforced by a process-wide `AgentGate` (Mutex+Condvar, RAII slot released on drop/panic). The gate bounds *execution*, not thread creation — manifests, task claims, and `Agent()` return semantics are unchanged. `TeamDelete` forgets the gate so queued agents drain. Gate regression test covers block-until-release and drop-releases-slot.
- `ToolSearch` deliberately stays sequential: the sequential path merges live MCP state; the stateless registry path wouldn't produce identical output.

**3. Workspace-scope check runs per call — confirmed**
Enforcement is inside each `GlobalToolRegistry::execute` arm (`classify_read_path_permission` + `maybe_enforce_permission_check_with_mode`, then `read_file_in_workspace` → `validate_workspace_boundary` against the workspace root per invocation). Nothing caches a scope decision for the batch; the `allowed_tools` guard also runs per call before dispatch. The ordering test above proves it empirically: an out-of-scope `../escape.txt` smuggled between five valid reads fails with `escapes workspace boundary` while its neighbors succeed.

Validation: `cargo fmt` clean; full test suites pass — runtime 615, tools 116, cli 212 (+integration suites); clippy on the three crates adds zero new warnings beyond the pre-existing baseline.

1716775457damn:
Thanks for the thorough follow-up — all three points are properly closed:

1. Panic-safe slot fallback for result ordering looks right; keying every spawn to its own (index, tool_use_id, tool_name) tuple removes the worst failure mode, and the workspace-escape regression test at cap=2 is a solid way to prove per-call ordering.
2. The AgentGate RAII bound is a clean choice — bounding execution rather than thread creation keeps manifest/task-claim semantics intact, and TeamDelete forgetting the gate so queued agents drain is a good edge-case catch.
3. Confirming the workspace-scope check runs per invocation (with allowed_tools still gating before dispatch) resolves my last concern about smuggling an out-of-scope path into a parallel batch.

This is ready once #3248 lands and the base commits are rebased away. No further blocking feedback from my side.

TheArchitectit:
Resubmitted as requested (re: [#3248 comment](https://github.com/ultraworkers/claw-code/pull/3248#issuecomment-5615253459)): rebased onto current `main` and dropped the three custom-OpenAI namespace commits (fefa608c, 2d6540c5, 0c339e7b) that #3248 supersedes. The PR now carries the 8 team commits only (head 77a23db).

Review note: with the trio dropped, the test `config_model_normalizes_bare_name_to_custom_prefix_for_custom_openai_provider` was also removed — it exercises the `custom/` model normalization that lands with #3248, and the intermediate push briefly failed CI on it. It should be restored and re-validated together with #3248's merge.

FYI: #3248 still shows OPEN / no merge commit on main as of 13:18 UTC on our side, so the dropped trio is not yet in main — nothing further needed from this PR when it lands.

1716775457damn:
Thanks for the cleanup — splitting the three OpenAI-compat namespace commits out and keeping this PR focused on the 8 team commits makes review much easier, especially once #3248 lands. The deterministic per-call ordering (panic-safe slot fallback) and the cap=2 ordering regression test look solid. Will re-review head-to-head and merge after CI passes.

1716775457damn:
Confirmed head 77a23db now carries just the 8 team commits — much cleaner to review than the previous 11-commit history. Alignment on the handoff: merge #3248 first, then restore the custom/ model-normalization test (config_model_normalizes_bare_name_to_custom_prefix_for_custom_openai_provider) and validate it against the merged base so we don't lose that coverage, then this PR gets its final pass. Once CI on the cleaned head is green, it's ready to merge from my side.

1716775457damn:
#3248 已合并进 main，之前的 merge-order 阻塞正式解除。rebase 掉三个 base commit 后请把 config_model_normalizes_bare_name_to_custom_prefix_for_custom_openai_provider 测试恢复并跑一遍（它验证 custom/ 前缀归一化，属于 #3248 落地的关键覆盖）。77a23db 恢复测试 + CI 全绿后我做最终 head-to-head 复评，通过即可合并。

1716775457damn:
状态同步：#3248 已合入 main，之前的 merge-order 阻塞已解除。请按上一条安排：rebase 后恢复 config_model_normalizes_bare_name_to_custom_prefix_for_custom_openai_provider 测试并确认通过（它覆盖 custom/ 前缀归一化，是 #3248 落地的关键校验），CI 全绿后我直接做最终 head-to-head 复评并合并。若 rebase 有冲突也可以贴出来一起看。

PR #3255, opened 2026-06-21
feat(cli): press Esc to interrupt a running turn

### LINKED ISSUE #3196: Feature Request: Escape key to interrupt a running agent turn
## Summary
  Currently, the only way to stop the harness mid-run is Ctrl+C, which kills the entire process.
  Claude Code supports pressing **Escape** to gracefully interrupt an in-progress agent turn
  without exiting the session.

  ## Current behavior

  - `input.rs` binds Ctrl+C → `ReadOutcome::Cancel` / `ReadOutcome::Exit`
  - No `KeyCode::Esc` binding exists during an active turn
  - No in-flight request cancellation is wired up

  ## Desired behavior

  Press **Escape** while the agent is running to:
  1. Cancel the active API request
  2. Transition the worker state from `Running` → `Interrupted`
  3. Return control to the REPL prompt without exiting the session

  ## Implementation sketch

  - Detect `KeyCode::Esc` in `rusty-claude-cli/src/input.rs` during an active turn
  - Add an `Interrupted` state to the worker lifecycle in `runtime/`
  - Cancel the in-flight HTTP request (e.g., via a `CancellationToken`)
  - Emit a cancellation event for observability

amazd:
## Summary

Press **Esc** to gracefully interrupt a running turn and return to the REPL prompt with the session intact. Previously the only way to stop a long-running turn was Ctrl+C, which left the in-flight request running.

This lands the feature in three layers:

1. **`TurnInterruptSignal` (runtime)** — a shared flag the conversation loop polls at safe points (before each model request, on stream failure, before each pending tool). When set, the turn winds down instead of failing: pending `tool_use`s get synthesized error `tool_result`s so every tool call stays answered and the session remains valid for the next request. `TurnSummary` gains an `interrupted` field and the tracer records a `turn_interrupted` event.

2. **In-flight request cancellation (CLI)** — `AnthropicRuntimeClient` races the streaming request against the interrupt flag with `tokio::select!`; tripping the flag drops the request future, aborting the HTTP stream for every provider. The per-turn Ctrl+C monitor now interrupts the whole turn rather than only aborting hooks.

3. **Esc key listener (CLI)** — `EscapeInterruptMonitor` switches stdin to a minimal non-canonical mode (echo/line-buffering off; `ISIG`/`OPOST` untouched, unlike full raw mode) for the duration of a turn and watches for a lone ESC byte. Arrow/function-key escape sequences are recognized and swallowed. Permission prompts that read whole lines mid-turn coordinate stdin ownership via `StdinPromptGate`, which restores canonical mode while a prompt holds a lease.

The spinner shows `(esc to interrupt)` and `WorkerStatus`/`WorkerEventKind` gain an `interrupted` lifecycle state. Non-terminal stdin and non-unix platforms skip the listener; Ctrl+C interruption still works there.

Closes #3196.

## Testing

- `scripts/fmt.sh --check` — clean
- Runtime: `interrupt_before_first_request_skips_the_api_call`, `interrupt_after_stream_synthesizes_results_for_pending_tools`, `stream_error_during_interrupt_is_reported_as_interruption`, `interrupted_status_serializes_and_displays_as_snake_case` — pass
- CLI: escape-byte classification unit tests (lone ESC vs. escape sequence vs. other) — pass

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2247,6 +2247,7 @@ dependencies = [
  "crossterm",
  "log",
  "mock-anthropic-service",
+ "nix",
  "plugins",
  "pulldown-cmark",
  "runtime",
--- a/rust/crates/runtime/src/conversation.rs
+++ b/rust/crates/runtime/src/conversation.rs
@@ -1,5 +1,9 @@
 use std::collections::BTreeMap;
 use std::fmt::{Display, Formatter};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
 
 use serde_json::{Map, Value};
 use telemetry::SessionTracer;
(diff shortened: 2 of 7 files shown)


DereC4:
image

1716775457damn:
Nice work! The three-layer approach (runtime flag, in-flight cancellation via tokio::select!, and CLI integration) is clean and well-structured. Esc is the natural key for this — much better than having to Ctrl+C and lose session state. The synthesized error tool_results for pending tool_uses is a thoughtful touch that keeps the turn consistent.

1716775457damn:
Much-needed UX improvement — Esc to interrupt is table-stakes for any interactive CLI. The crossterm event loop integration looks clean. Thanks @amazd!

1716775457damn:
Nice, the `TurnInterruptSignal` approach with synthesized error tool_results is a solid design choice — keeping sessions valid across interruptions avoids a lot of the pain from Ctrl+C use cases. One thing I'd be curious about: when the interrupt fires mid-tool-execution (say a long-running `shell_executor` call), does the synthesized `tool_result` propagate a distinct error code (e.g. `interrupted`) so the model can distinguish an interrupt from a genuine tool failure? That could be useful for the model to adjust its next-turn strategy. Otherwise this looks great.

1716775457damn:
Esc to interrupt is a natural UX expectation coming from terminal workflows. This removes friction when the model is going down a wrong path.

1716775457damn:
Two things worth nailing down before merge, since interrupt semantics tend to leak: (1) the signal is polled at safe points, so a long-running tool subprocess will not stop until it returns — is there a paired kill/timeout for the child, or does Esc only take effect between steps? Worth documenting which phases are actually interruptible. (2) Raw-mode terminals send Esc as the first byte of many escape sequences (arrow keys, Alt+key), so the reader needs to distinguish a bare Esc from a sequence prefix, usually with a short timeout before treating it as standalone; otherwise arrow-key input will cancel turns. A test that feeds a partial escape sequence would cover it.

PR #3263, opened 2026-06-25
Improve command lookup normalization

hiSandog:
## Summary
- Trim and normalize command lookup input before alias resolution.
- Prioritize exact command-name matches before broader substring matches.
- Add coverage for whitespace/case-normalized command lookup and execution.

## Validation
- env PYTHONDONTWRITEBYTECODE=1 python3 -m unittest tests.test_porting_workspace.PortingWorkspaceTests.test_commands_and_tools_cli_run tests.test_porting_workspace.PortingWorkspaceTests.test_command_lookup_normalizes_user_input_whitespace
- git diff --check

### DIFF
--- a/src/commands.py
+++ b/src/commands.py
@@ -55,7 +55,8 @@ def command_names() -> list[str]:
 
 
 def get_command(name: str) -> PortingModule | None:
-    needle = COMMAND_ALIASES.get(name.lower(), name.lower())
+    normalized = name.strip().lower()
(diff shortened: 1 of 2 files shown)


1716775457damn:
Good improvement! Trim + normalize before alias resolution is a sensible approach, and prioritizing exact matches over substring matches avoids ambiguity. Test coverage for whitespace/case normalization looks solid.

PR #3232, opened 2026-06-06
fix(providers): strip provider prefix from model names for openai_com…

Ajinkya-Ghuge:
…pat endpoints

## Summary
Fixes an issue where OpenAI-compatible clients (like DeepSeek) fail due to strict provider/model naming conventions by dynamically stripping the provider prefix before sending the payload downstream.

## Anti-slop triage
- Classification: 
- Evidence: 
- Non-destructive review result: 

## Verification
- [ ] Targeted tests/docs checks ran, or the gap is explicitly recorded.
- [ ] `git diff --check` passes.
- [ ] No live secrets, tokens, private logs, or unrelated generated churn are included.

## Resolution gate
Closes #3229
- [ ] If this PR resolves an issue, the issue number and fix evidence are linked.
- [ ] If this PR should not merge, the rejection/defer rationale is evidence-backed and does not rely on vibes.
- [ ] I did not merge/close remote PRs or issues from an automation lane without owner approval.

PR #3268, opened 2026-07-09
docs(rust/README): document xAI and Kimi model aliases

nankingjing:
## What

The **Model Aliases** table in `rust/README.md` only listed the three Anthropic
aliases (`opus`, `sonnet`, `haiku`), but `resolve_model_alias` in
`rust/crates/api/src/providers/mod.rs` also resolves xAI and Kimi aliases. This
PR documents the full alias set, adds a **Provider** and **Auth env var** column
so each alias is actionable, and notes that resolution is case-insensitive.

Newly documented aliases (verified against `MODEL_REGISTRY` / `resolve_model_alias`):

| Alias | Resolves To | Provider | Auth env var |
|-------|-------------|----------|--------------|
| `grok`, `grok-3` | `grok-3` | xAI | `XAI_API_KEY` |
| `grok-mini`, `grok-3-mini` | `grok-3-mini` | xAI | `XAI_API_KEY` |
| `grok-2` | `grok-2` | xAI | `XAI_API_KEY` |
| `kimi` | `kimi-k2.5` | DashScope (OpenAI-compatible) | `DASHSCOPE_API_KEY` |

## Why it's correct

Everything added is verifiable in the source, not invented:

- `resolve_model_alias()` maps `grok`/`grok-3` → `grok-3`, `grok-mini`/`grok-3-mini`
  → `grok-3-mini`, `grok-2` → `grok-2`, and `kimi` → `kimi-k2.5`.
- `MODEL_REGISTRY` sets the auth/base-url env vars for each provider:
  xAI → `XAI_API_KEY` / `XAI_BASE_URL`, Kimi → `DASHSCOPE_API_KEY` /
  `DASHSCOPE_BASE_URL`.
- Case-insensitivity: `resolve_model_alias` lowercases input before matching, and
  existing tests assert this (`kimi_alias_resolves_to_kimi_k2_5` checks `"KIMI"`,
  `resolves_grok_aliases` checks the grok family).
- Provider diagnostics already point users at these aliases (e.g. the hint to
  "use an xAI model alias such as `--model grok`"), and `docs/MODEL_COMPATIBILITY.md`
  documents the corresponding request handling — this PR just surfaces the aliases
  where they are listed.

## Scope

Docs-only change to `rust/README.md`. No code or behavior changes. The existing
`opus`/`sonnet`/`haiku` rows are unchanged.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -110,13 +110,27 @@ Primary artifacts:
 
 ## Model Aliases
 
-Short names resolve to the latest model versions:
+Short model aliases resolve to canonical model IDs. Because the router selects a
+provider from the resolved model, an alias also determines which backend and
+credentials are used — choose the alias for the provider you have an API key for.
+Alias resolution is case-insensitive (`OPUS` behaves the same as `opus`).
 
-| Alias | Resolves To |
-|-------|------------|
-| `opus` | `claude-opus-4-7` |
-| `sonnet` | `claude-sonnet-4-6` |
-| `haiku` | `claude-haiku-4-5-20251213` |
+| Alias | Resolves To | Provider | Auth env var |
+|-------|-------------|----------|--------------|
+| `opus` | `claude-opus-4-7` | Anthropic | `ANTHROPIC_API_KEY` |
+| `sonnet` | `claude-sonnet-4-6` | Anthropic | `ANTHROPIC_API_KEY` |
+| `haiku` | `claude-haiku-4-5-20251213` | Anthropic | `ANTHROPIC_API_KEY` |
+| `grok`, `grok-3` | `grok-3` | xAI | `XAI_API_KEY` |
+| `grok-mini`, `grok-3-mini` | `grok-3-mini` | xAI | `XAI_API_KEY` |
+| `grok-2` | `grok-2` | xAI | `XAI_API_KEY` |
+| `kimi` | `kimi-k2.5` | DashScope (OpenAI-compatible) | `DASHSCOPE_API_KEY` |
+
+Anthropic aliases track the latest published model versions. xAI models honor
+`XAI_BASE_URL` and Kimi honors `DASHSCOPE_BASE_URL` for base-URL overrides. Alias
+resolution is implemented in `resolve_model_alias`
+([`crates/api/src/providers/mod.rs`](./crates/api/src/providers/mod.rs)); for
+provider-specific request handling see
+[`../docs/MODEL_COMPATIBILITY.md`](../docs/MODEL_COMPATIBILITY.md).
 
 ## CLI Flags and Commands
 


1716775457damn:
Good catch — the model aliases table was indeed incomplete. Having the full alias set documented alongside the Provider and Auth env var columns makes it much more actionable for users. The case-insensitivity note is also a helpful detail.

simawulei:
我已收到您的邮件~会尽快给您回复。祝您六时吉祥，阿弥陀佛。

nankingjing:
Thanks for the quick review, @1716775457damn! Glad the Provider / Auth-env columns are useful. I noticed a couple of other small doc nits while in there (a broken link and an out-of-date crate count) and have a follow-up PR incoming, kept separate to keep this one focused.

1716775457damn:
Sounds good, keeping the follow-up PR focused on those doc nits separately makes sense. Looking forward to reviewing it — especially the broken link fix. Thanks for the thorough docs work!

nankingjing:
Quick note: #3269 is just the crate-count fix; I haven't opened the broken-link follow-up yet — will get to that separately. Thanks for flagging it!

1716775457damn:
Got it, thanks for the clarification. #3269 looks good for the crate count — will review that next. No rush on the broken link, just drop a mention when it's up.

1716775457damn:
The alias table now covers all provider models consistently — having xAI and Kimi documented alongside Anthropic makes it much easier to discover available options. Thanks for filling this gap @nankingjing!

1716775457damn:
Thanks for the update @nankingjing! The alias table looks complete now — covers xAI and Kimi models cleanly. Docs-only change, good to go.

nankingjing:

1716775457damn:
Approved. Filling the gap where xAI and Kimi aliases were missing from the documented table while already supported in resolve_model_alias is the right fix. The additional Provider and Auth env var columns make each alias immediately actionable rather than requiring users to cross-reference config files. Good docs improvement.

1716775457damn:
Useful addition. xAI and Kimi are increasingly popular, documenting their aliases helps users get started quickly.

1716775457damn:
One follow-up worth considering: this drift happened because the alias table is maintained by hand while the real resolution lives in resolve_model_alias, so the same gap reopens the next time an alias is added. Generating the table from the resolver, or adding a small test asserting that every alias listed in the README actually resolves (and that every resolver alias is documented), would keep docs and behaviour from diverging. It would also help to state whether alias resolution happens before or after the provider-prefix stripping from #3232, since that ordering is what users will hit.

nankingjing:

The head commit (`b89ddda387`) has **0 check runs and 0 workflow runs** — CI has never run on this PR at all. `commits/b89ddda387/status` reports `pending` with 0 statuses, which is an absence of results rather than a pass.

nankingjing:
Answering the follow-up from @1716775457damn (2026-09-07) properly, rather than only the CI note above.

**The drift is real, and slightly wider than this diff implies.** In `rust/crates/api/src/providers/mod.rs`, `MODEL_REGISTRY` (line 121) holds nine aliases: `opus`, `sonnet`, `haiku`, `grok`, `grok-3`, `grok-mini`, `grok-3-mini`, `grok-2`, `kimi`. The README documented three — the `## Model Aliases` table is three rows, and the feature matrix row says "Model aliases (opus/sonnet/haiku)". The six undocumented ones are exactly the six this PR adds, so the table is now complete; you're right that nothing keeps it that way.

**On generating the table.** I'd rather land a parity test than promise a generator, and it's worth being precise about why the generator is not trivial: the alias→canonical-id mapping is *not* in `MODEL_REGISTRY`. It is a second, separate `match` inside `resolve_model_alias` (`providers/mod.rs:206`) — `ProviderKind::Anthropic => match *alias { "opus" => "claude-opus-4-7", ... }`, and likewise for Xai and OpenAi. So a generator would have to read both structures, and would still be guessing which one owns the mapping. The cleaner fix is to move the canonical id onto the registry entry so `resolve_model_alias` becomes a lookup and the README row is a projection of one structure. That is a refactor of the resolver, not of the docs, so I'd rather not smuggle it into a docs PR — happy to open it separately if you want it. A parity test, by contrast, I can add here and it costs nothing.

**On the ordering question (alias resolution vs the provider-prefix stripping).** I traced it: the two do not compete, and the order is alias-first, with prefix handling only ever seeing the already-resolved name.

- `rust/crates/api/src/client.rs:25` — `let resolved_model = providers::resolve_model_alias(model);`, and `detect_provider_kind` is then called on the *result*.
- `detect_provider_kind` (`providers/mod.rs:362`) also calls `resolve_model_alias` itself, then delegates to `metadata_for_model`.
- `metadata_for_model` (`providers/mod.rs:235`) is where prefixes are inspected — `anthropic/`, `openai/`, `gpt-`, `local/` — each a `starts_with` on the canonical string.

Two consequences worth stating for users:

1. Alias matching is a **whole-string** equality test against the registry, so `anthropic/opus` is not an alias and won't become one; it routes through the `anthropic/` prefix branch instead. Every alias this PR documents is a bare name.
2. The only place a provider prefix is actually *stripped* is downstream of routing, in the Anthropic client — `anthropic_wire_model` at `providers/anthropic.rs:1038`, `model.strip_prefix("anthropic/")` — applied when the request is built. So the pipeline is: alias resolves → prefix decides routing → strip happens at request build.

Tell me if you want the parity test in this PR or as a follow-up; either is fine by me.

1716775457damn:
Parity test in this PR, please — it's cheap and it locks the exact property that drifted, so there's no reason to defer it. Agreed on keeping the registry refactor out: moving the canonical id onto the registry entry is a resolver change, and doing it as its own PR keeps this docs fix reviewable. Thanks for tracing the ordering — alias-first with prefix stripping only at request-build time is worth one line in the README so users don't have to guess whether "anthropic/opus" is a valid alias (it isn't).

nankingjing:
Done — the parity test is in (`072594d`), with the README line alongside it.

What it locks, in both directions:

- every alias in `MODEL_REGISTRY` appears in the `## Model Aliases` table, carrying the canonical id it actually resolves to and the auth env var its provider metadata reports;
- the table introduces no alias the resolver does not know — it is set equality, so a stale row fails as loudly as a missing one;
- each row is re-checked uppercased, since the README claims `OPUS` behaves the same as `opus`.

It parses the README rather than hardcoding expectations, so editing the table without the registry (or the reverse) fails with the offending row named in the assertion message.

One deviation worth flagging, since it changes the wording you suggested. I wrote the README line as "a prefixed form such as `anthropic/opus` is not an alias, and is passed through as-is rather than expanded to `claude-opus-4-7`" instead of describing prefix stripping at request-build time. `strip_provider_prefix` in `providers/mod.rs` has no call sites anywhere in the workspace — it is currently dead code. The slash handling that does run at request-build time (`rsplit('/').next()` in `model_token_limit` and the `openai_compat` paths) takes the *suffix*, so "stripping" is not what happens to a prefixed alias today. The user-visible claim is the same either way — `anthropic/opus` does not expand to `claude-opus-4-7` — and that is the property the test asserts, so I documented the behaviour rather than the mechanism. Happy to reword to the mechanism if you would rather the README describe the intent, though that reads to me like it would promise something the code does not yet do.

`cargo test -p api` is green locally: 160 passed, 0 failed. `cargo fmt --check` and `cargo clippy -p api` are clean with no new warnings.

Worth knowing for this and the other PRs: the push created two workflow runs (`Rust`, `Rust CI`) and both are sitting at `action_required` with zero check runs, so nothing will show up on CI until a maintainer approves the runs for the new head commit.

1716775457damn:
Parity test 的设计很到位：解析 README 而非硬编码期望，集合相等让 stale row 和缺失行同样失败，大写复查也覆盖了 README 的声明。文档化行为而不是机制的取舍我赞成——写成 prefix stripping 反而会承诺 strip_provider_prefix 这个死代码做不到的事，现在这句 user-visible 声明才是测试真正断言的属性。CI 等维护者批准 action_required 即可，本地 160 passed 已经有足够信心。

PR #3269, opened 2026-07-10
docs: correct rust workspace crate count to 11

nankingjing:
## Summary

Follow-up doc fix after #3268.

The Stats section of `rust/README.md` states **9 crates** in the workspace, but the workspace (`members = ["crates/*"]`) now contains **11** crates:

`api`, `claw-analog`, `claw-rag-service`, `commands`, `compat-harness`, `mock-anthropic-service`, `plugins`, `runtime`, `rusty-claude-cli`, `telemetry`, `tools`

(verified by counting `rust/crates/*/Cargo.toml` — 11 manifests). This PR updates only that number; it does not restructure the layout or responsibilities sections.

## Note on a second candidate fix (not included)

While investigating I also checked the reported broken link in `how_to_run.md` (`[futute.md](futute.md)`). I did **not** change it: neither `futute.md` nor a correctly-spelled `future.md` exists anywhere in the repository, so retargeting the link would not resolve to a real file. Left untouched pending a decision on whether that target doc should be created.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -221,7 +221,7 @@ rust/
 ## Stats
 
 - **~20K lines** of Rust
-- **9 crates** in workspace
+- **11 crates** in workspace
 - **Binary name:** `claw`
 - **Default model:** `claude-opus-4-7`
 - **Default permissions:** `workspace-write`


1716775457damn:
Good catch — docs that say one thing while the workspace defines another are a real source of confusion for new contributors. With 11 crates now, are you planning to document what the 11th is separately or leave it self-explanatory from the workspace layout?

nankingjing:
The workspace uses `members = ["crates/*"]` so any directory dropped under `rust/crates/` automatically becomes a member — there isn't really a single 11th crate, the count just drifted from 9 to 11 over time. All 11 are listed in the PR description (api, claw-analog, claw-rag-service, commands, compat-harness, mock-anthropic-service, plugins, runtime, rusty-claude-cli, telemetry, tools). I figured that plus the glob pattern makes it self-explanatory. Happy to add a one-liner listing them in the README if you think it'd help though.

1716775457damn:
Straightforward doc fix, the drift from 9 to 11 is exactly the kind of thing that silently causes confusion. Since the workspace uses `members = ["crates/*"]` the count is inherently dynamic, but having an accurate number in the README still helps set expectations. I think the glob pattern makes it self-explanatory enough — no need to list all 11 individually unless someone asks.

nankingjing:
Thanks for the confirmation — makes sense to keep it lean without listing all 11. I'll leave it as-is.

1716775457damn:
Confirmed — keeping it lean without enumerating all 11 crates is the right call. The glob pattern handles membership dynamically; the accurate count in the README is enough to set expectations. Simple and effective fix.

1716775457damn:
Confirmed — the count was clearly stale (9 vs the actual 11). Good call to keep it as a single number rather than enumerate, since the workspace already uses a glob. Nice clean follow-up to #3268.

1716775457damn:
Good catch — workspace now has 11 crates. Simple doc fix, LGTM. Thanks @nankingjing!

nankingjing:

1716775457damn:
Approved. Simple doc fix — 9→11 is exactly right given the members=['crates/*'] glob. Keeping the count as a single number rather than enumerating all crates is the right call, since the glob handles membership dynamically. Clean follow-up to #3268.

1716775457damn:
Already approved. Simple doc fix — 9 to 11 crates, verified against the workspace members=['crates/*'] glob. Keeping it as a number rather than enumerating is the right call since membership is dynamic.

1716775457damn:
Good fix, 11 crates is correct. Inaccurate docs can mislead contributors.

nankingjing:

The head commit (`8613b7065a`) has **0 check runs and 0 workflow runs** — CI has never run on this PR at all. `commits/8613b7065a/status` reports `pending` with 0 statuses, which is an absence of results rather than a pass.

1716775457damn:
Appreciate the correction — nothing to apologize for. With 0 check runs on the head commit there was never a real CI signal, and since both of us are outside contributors my approval doesn't unblock anything anyway. Nothing further for me to do here; the PR needs a maintainer to trigger the held workflow runs and merge. The count fix itself is sound and I stand by the review.

PR #3271, opened 2026-07-10
fix: generalize local-model validation bypass beyond Ollama tags

Aneesh-382005:
## Summary
- `is_local_openai_model_syntax()` only bypassed strict `provider/model` validation when the model name contained a `.` or `:` (Ollama-tag-shaped). Bare names LM Studio/llama.cpp commonly use (e.g. `mistral`) were rejected unless prefixed with `local/`.
- Fix: reuse the existing `is_local_openai_compatible_base_url()` check (already used to waive the API-key requirement) to also waive the model-name shape check when `OPENAI_BASE_URL` is loopback/private. Remote gateways keep the old heuristic, so a mistyped bare name can't silently route to a paid API.
- Scoped narrowly on purpose **see Future plans.**

## Anti-slop triage
- Evidence: `OPENAI_BASE_URL=http://127.0.0.1:1234/v1 claw --model "mistral" prompt "hi"` fails today with `invalid model syntax` despite the server serving that model. Verified fixed against a local mock server; verified remote gateways (`openrouter.ai`) still reject bare names.

## Verification
- [x] `scripts/fmt.sh --check`, `cargo test --workspace` all pass, incl. 4 new tests.
- [x] `git diff --check` passes.
- [x] No secrets/unrelated churn; diff scoped to 6 files.
- Note: `cargo clippy --workspace -- -D warnings` fails, but identically on unmodified `main` (confirmed via `git stash`) -> pre-existing, not touched by this PR.

## Resolution gate
- [x] Follow-up to my earlier PR #3213 (native `OLLAMA_HOST` support); ROADMAP entry added.
- [ ] N/A - intended to merge.
- [x] N/A - no remote PR/issue automation actions taken.

## Future plans
Deferring two follow-ups raised in #3213's review to separate PRs: named per-backend env vars (`LMSTUDIO_HOST`, etc.) and a config-based local-provider registry. Keeping this one small and reviewable.


1716775457damn:
Good fix — the restriction that only Ollama-style tags (with . or :) could bypass validation was a real pain point for LM Studio/llama.cpp users with bare model names. Reusing the existing is_local_openai_compatible_base_url check to also waive the model-name shape constraint is the right approach — it's scoped to loopback/private URLs, so remote gateways keep validation. The anti-slop triage section with the concrete reproducer is appreciated.

Aneesh-382005:
> Good fix — the restriction that only Ollama-style tags (with . or :) could bypass validation was a real pain point for LM Studio/llama.cpp users with bare model names. Reusing the existing is_local_openai_compatible_base_url check to also waive the model-name shape constraint is the right approach — it's scoped to loopback/private URLs, so remote gateways keep validation. The anti-slop triage section with the concrete reproducer is appreciated.

Thanks for the review! Once this merges, I'm planning a couple of follow-up PRs for named per-backend environment variables and a config-based local provider registry. Do you think it's better to wait for this to merge before opening those, or would it be fine to work on them in parallel?

1716775457damn:
Reusing is_local_openai_compatible_base_url to gate the bypass is a clean approach — it keeps the Ollama heuristic for remote gateways while removing the friction for local setups. The scope is appropriately narrow; no risk of accidentally routing bare names to paid APIs.

1716775457damn:
Good generalization — the old check was too tightly coupled to Ollama tag format. The model name heuristic approach covers local models from any provider. Thanks @Aneesh-382005!

1716775457damn:
This is a well-scoped fix. The key insight — that is_local_openai_compatible_base_url() already gates the API-key waiver and should also gate the model-name shape check — avoids duplication and keeps the security boundary clear (remote gateways retain strict validation). The 4 new tests and the decision to defer per-backend env vars to a separate PR keeps this focused and reviewable. The clippy failure being pre-existing (confirmed on main) is also good hygiene to document upfront.

1716775457damn:
Good catch. The local-model validation bypass created a security gap. Generalizing the fix across all model validation paths is the right approach.

1716775457damn:
Regarding your follow-up question: I'd suggest getting this PR merged first before opening the named per-backend env vars and provider registry follow-ups. They'll likely touch the same validation/config area, so landing this one first keeps the diffs independent and avoids review churn. If those follow-ups are independent enough, opening them as drafts in parallel is also fine — but sequencing after merge is cleaner.

PR #3272, opened 2026-07-11
fix(cli): preserve trailing block separators in streamed markdown

### LINKED ISSUE #3257: bug: Missing newlines before paragraphs and code blocks in streamed Markdown rendering
### Describe the bug

When streaming responses from an OpenAI-compatible API (such as Grok or custom proxy gateways), paragraphs and code blocks sometimes lose their leading and trailing newlines, causing them to merge directly into the previous output line without proper line breaks in the terminal.

For example, a response might render as:
`最简单的Python入门例子：Hello World！1. 打开记事本` (instead of having double newlines before `1.`)
And:
`粘贴进去（全部内容就是它）：╭─ python` (instead of having double newlines before the code block)

### Root Cause Analysis

The bug is located in `TerminalRenderer::render_markdown` inside `rust/crates/rusty-claude-cli/src/render.rs`:

```rust
    #[must_use]
    pub fn render_markdown(&self, markdown: &str) -> String {
        let normalized = normalize_nested_fences(markdown);
        let mut output = String::new();
        // ...
        for event in Parser::new_ext(&normalized, Options::all()) {
            self.render_event(...);
        }

        output.trim_end().to_string() // <--- HERE
    }
```

During streaming, `MarkdownStreamState::push` parses intermediate chunks (split at boundaries using `find_stream_safe_boundary`).
- The `ready` chunk (e.g., up to an empty line representing a paragraph break) correctly ends with trailing newlines (e.g., `\n\n` generated by paragraph boundaries).
- However, `renderer.markdown_to_ansi(&ready)` calls `render_markdown()`, which unconditionally performs `.trim_end()` on the rendered string.
- This strips all trailing newlines (including the structural `\n\n` representing paragraph/block separators).
- Since these newlines are stripped from the output chunk and the original `ready` text has been drained from `self.pending`, the newlines are completely lost from the terminal print.
- When the next chunk (e.g. list items or code block starts) is printed, it is outputted directly appended to the previous text without any line breaks.

### Suggested Fix

Introduce a stream-friendly rendering method in `TerminalRenderer` that preserves trailing whitespace/newlines during stream chunk generation, and use it inside `MarkdownStreamState::push`.

For example:

1. In `TerminalRenderer` (inside `rust/crates/rusty-claude-cli/src/render.rs`):
```rust
    #[must_use]
    pub fn render_markdown_stream(&self, markdown: &str) -> String {
        let normalized = normalize_nested_fences(markdown);
        let mut output = String::new();
        let mut state = RenderState::default();
        let mut code_language = String::new();
        let mut code_buffer = String::new();
        let mut in_code_block = false;

        for event in Parser::new_ext(&normalized, Options::all()) {
            self.render_event(
                event,
                &mut state,
                &mut output,
                &mut code_buffer,
                &mut code_language,
                &mut in_code_block,
            );
        }

        output // Do not call trim_end() here
    }

    #[must_use]
    pub fn markdown_to_ansi_stream(&self, markdown: &str) -> String {
        self.render_markdown_stream(markdown)
    }
```

2. In `MarkdownStreamState::push`:
```rust
    #[must_use]
    pub fn push(&mut self, renderer: &TerminalRenderer, delta: &str) -> Option<String> {
        self.pending.push_str(delta);
        let split = find_stream_safe_boundary(&self.pending)?;
        let ready = self.pending[..split].to_string();
        self.pending.drain(..split);
        Some(renderer.markdown_to_ansi_stream(&ready)) // Use the stream-specific method
    }
```

This keeps `MarkdownStreamState::flush` using the standard `markdown_to_ansi` (which trims trailing newlines at the very end of the stream), while correctly outputting structural paragraph and code block newlines during streaming.

nankingjing:
## Summary

Fixes #3257. During streaming, block-separating newlines were being dropped, causing paragraphs, lists and code blocks to run into the previous line in the terminal (e.g. `...Hello World！1. 打开记事本` and `...：╭─ python`).

## Root cause

`MarkdownStreamState::push` renders each ready chunk with `TerminalRenderer::markdown_to_ansi`, which calls `render_markdown` → `output.trim_end()`. That trim strips the structural trailing `\n\n` that the parser emits at the end of a paragraph/heading/list/code block. In `run_turn` the streamed chunk is written straight to the terminal with `write!(out, "{rendered}")` and **no** separator of its own (see `main.rs`), so once the trailing newlines are trimmed the next block is printed flush against the previous one.

The final `flush` path legitimately wants the tail trimmed, so only the per-chunk `push` path is wrong.

## Fix

- Extract the untrimmed render into a private `render_markdown_raw`; `render_markdown` now trims its result (behavior unchanged for one-shot rendering).
- Add `markdown_to_ansi_stream`, a stream-friendly variant that returns the untrimmed output.
- `MarkdownStreamState::push` now uses `markdown_to_ansi_stream`, preserving the block separators between chunks. `flush` is unchanged and still trims the final tail.

Because `pulldown-cmark` collapses runs of blank lines and chunk boundaries fall on blank lines, each block still ends with at most one blank line — no extra blank-line accumulation is introduced.

## Test

Adds `stream_rendering_preserves_block_separators`, which asserts the stream variant keeps trailing newlines (while `markdown_to_ansi` still trims) and that two consecutive streamed chunks do not stick together.

## Verification

Verified by reading the source and tracing the parser/stream logic; **not compiled or executed** in this environment. The change is additive plus a one-line call swap, matches the existing `render_markdown`/`markdown_to_ansi` convention, and preserves all existing tests.

### DIFF
--- a/rust/crates/rusty-claude-cli/src/render.rs
+++ b/rust/crates/rusty-claude-cli/src/render.rs
@@ -249,6 +249,16 @@ impl TerminalRenderer {
 
     #[must_use]
     pub fn render_markdown(&self, markdown: &str) -> String {
+        self.render_markdown_raw(markdown).trim_end().to_string()
+    }
+
+    /// Render markdown to ANSI without trimming trailing whitespace.
+    ///
+    /// Block-level events (paragraphs, headings, lists, code blocks) emit their
+    /// trailing newline separators here. `render_markdown` trims them for
+    /// one-shot output, while streaming keeps them so that successive chunks
+    /// stay on their own lines instead of running together.
+    fn render_markdown_raw(&self, markdown: &str) -> String {
         let normalized = normalize_nested_fences(markdown);
         let mut output = String::new();
         let mut state = RenderState::default();
@@ -267,12 +277,24 @@ impl TerminalRenderer {
             );
         }
 
-        output.trim_end().to_string()
+        output
     }
 
     #[must_use]
     pub fn markdown_to_ansi(&self, markdown: &str) -> String {
         self.render_markdown(markdown)
+    }
+
+    /// Stream-friendly variant of `markdown_to_ansi` that preserves trailing
+    /// block separators (such as a paragraph's closing newlines).
+    ///
+    /// `MarkdownStreamState::push` writes each rendered chunk directly to the
+    /// terminal with no separator of its own, so the trailing newlines must
+    /// survive here to keep a paragraph from sticking to the block that follows
+    /// it. The final `flush` still uses `markdown_to_ansi` to trim the tail.
+    #[must_use]
+    pub fn markdown_to_ansi_stream(&self, markdown: &str) -> String {
+        self.render_markdown_raw(markdown)
     }
 
     #[allow(clippy::too_many_lines)]
@@ -609,7 +631,7 @@ impl MarkdownStreamState {
         let split = find_stream_safe_boundary(&self.pending)?;
         let ready = self.pending[..split].to_string();
         self.pending.drain(..split);
-        Some(renderer.markdown_to_ansi(&ready))
+        Some(renderer.markdown_to_ansi_stream(&ready))
     }
 
     #[must_use]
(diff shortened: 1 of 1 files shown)


1716775457damn:
Nice edge-case fix — trailing block separators getting stripped during streaming would break markdown rendering. The fix is minimal and correctly scoped. Thanks @nankingjing!

nankingjing:

1716775457damn:
Approved. trim_end() stripping structural trailing newlines during streaming is a subtle but impactful bug — paragraphs and code blocks bleeding together in the terminal would be confusing. The fix looks correctly scoped to preserve block separators without reintroducing unwanted trailing whitespace.

1716775457damn:
Already approved. The fix is correctly scoped — markdown_to_ansi_stream preserving block separators without changing the existing trim behavior for one-shot rendering is exactly the right approach.

1716775457damn:
Good analysis of the root cause in output.trim_end() — the distinction between one-shot rendering (where trimming is correct) and streamed chunks (where it drops structural separators) is precisely the right framing. The markdown_to_ansi_stream approach keeps the existing API intact while solving the streaming issue cleanly.

Will go through and submit formal approvals on the green PRs. Thanks for the fixes @nankingjing!

1716775457damn:
Good edge case fix. Trailing block separators being stripped during streaming could break markdown rendering of subsequent blocks.

nankingjing:

The head commit (`258dff2237`) has **0 check runs and 0 workflow runs** — CI has never run on this PR at all. `commits/258dff2237/status` reports `pending` with 0 statuses, which is an absence of results rather than a pass.

1716775457damn:
Thanks for the correction — no apology needed. Makes sense: with 0 check runs on the head commit there was never a real CI signal, and as an outside contributor my approval doesn't gate the held workflows anyway. Nothing for me to do on this one; the PR is waiting on a maintainer to approve the queued workflow runs and merge. The fix itself reads fine and I stand by the code review.

PR #3273, opened 2026-07-11
fix(cli): stop reprinting streamed assistant text after 'Done' in non-compact text mode

### LINKED ISSUE #3258: fix(cli): resolve duplicate assistant text printing after 'Done' in non-compact text mode
### Description
When running `claw` in non-compact text mode, the entire assistant response is printed twice.
1. First, the text chunks are streamed and printed in real-time during the API call.
2. Second, after `✔ ✨ Done` is printed, the full assistant text (`final_text`) is printed again.

This causes a large amount of redundant output in the console, especially for long responses.

### Cause
In `LiveCli::run_turn` (inside `main.rs`), `spinner.finish("✨ Done")` is called right after the conversation stream finishes. 
Because `spinner.finish` moves the cursor to column 0 and clears the current line, it would erase the last line of the streamed text if we do not print a newline first.
To prevent this, the previous commit introduced a secondary full-text reprint:
```rust
let final_text = final_assistant_text(&summary);
if !final_text.is_empty() {
    let renderer = TerminalRenderer::new();
    println!("{}", renderer.markdown_to_ansi(&final_text));
}
```
However, this reprints the *entire* assistant message instead of just protecting the last line, causing the duplicate response.

### Solution
1. Print a newline before `spinner.finish` if the `final_text` is not empty. This moves the cursor to a new line, ensuring the line-clearing operation in `spinner.finish` only clears the empty line and does not corrupt the last line of the streamed text.
2. Remove the redundant `final_text` reprint logic in `run_turn`.

Here is the proposed diff in `rust/crates/rusty-claude-cli/src/main.rs`:
```diff
@@ -7763,15 +7763,15 @@
         match result {
             Ok(summary) => {
                 self.replace_runtime(runtime)?;
-                spinner.finish(
-                    "✨ Done",
-                    TerminalRenderer::new().color_theme(),
-                    &mut stdout,
-                )?;
                 let final_text = final_assistant_text(&summary);
                 if !final_text.is_empty() {
-                    let renderer = TerminalRenderer::new();
-                    println!("{}", renderer.markdown_to_ansi(&final_text));
-                }
+                    println!();
+                }
+                spinner.finish(
+                    "✨ Done",
+                    TerminalRenderer::new().color_theme(),
+                    &mut stdout,
+                )?;
                 println!();
                 if let Some(event) = summary.auto_compaction {
```

nankingjing:
## Summary

Fixes #3258.

In **non-compact text mode** (and the interactive REPL), `LiveCli::run_turn`
prints the entire assistant response **twice**:

1. It runs the turn with `emit_output = true`, so `AnthropicRuntimeClient::consume_stream`
   already streams and renders every text delta to the terminal live
   (`main.rs`, `ContentBlockDelta::TextDelta` / `MessageStop`).
2. After `spinner.finish("✨ Done")` it then calls
   `println!("{final_text}")`, reprinting the whole message that was just streamed.

The reprint was originally added because `Spinner::finish` runs
`MoveToColumn(0)` + `Clear(ClearType::CurrentLine)` and would otherwise erase
the last streamed line (the streamed tail has no trailing newline). Reprinting
the full text "fixed" the erased line but duplicated everything above it.

## Fix

Protect only the last streamed line instead of reprinting the whole message:

- When there is streamed assistant text, emit a single `println!()` **before**
  `spinner.finish` so the line-clear lands on a fresh empty line rather than the
  last content line.
- Remove the `println!("{final_text}")` reprint.

`final_text` is still computed and used to decide whether a newline is needed,
so the empty-response path (tool-only turns) is unchanged. Compact/JSON paths
use `emit_output = false` and print the message exactly once — they are
untouched.

## Scope

- 1 file, `rust/crates/rusty-claude-cli/src/main.rs` (`LiveCli::run_turn`), +9 -4.
- No signature/API changes; no test depended on the duplicate output.

## Verification

Verified by reading and tracing the streaming path (`consume_stream` writes each
text delta live when `emit_output` is true; `prepare_turn_runtime(true)` is used
by `run_turn`, while the compact/JSON paths use `prepare_turn_runtime(false)`)
and by inspecting `Spinner::finish`'s line-clear behavior. **Not compiled or
executed** in this environment (full Rust workspace build not available here).
The change is a local reordering plus removal of one `println!`; `final_text`
remains used, so no unused-variable warning is introduced.

### DIFF
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -7908,6 +7908,9 @@ impl LiveCli {
                         match new_runtime.run_turn(input, Some(&mut rp)) {
                             Ok(summary) => {
                                 self.replace_runtime(new_runtime)?;
+                                if !final_assistant_text(&summary).is_empty() {
+                                    println!();
+                                }
                                 spinner.finish(
                                     if round == 0 {
                                         "✨ Done (after auto-compact)"
--- a/rust/crates/rusty-claude-cli/src/main.rs
+++ b/rust/crates/rusty-claude-cli/src/main.rs
@@ -7763,15 +7763,20 @@ impl LiveCli {
         match result {
             Ok(summary) => {
                 self.replace_runtime(runtime)?;
+                // The assistant response was already streamed to the terminal live
+                // during the turn (emit_output = true), so reprinting the full text
+                // here would duplicate the entire message. When text was streamed,
+                // emit a newline first so `spinner.finish`'s line-clear lands on a
+                // fresh line instead of erasing the last streamed line.
+                let final_text = final_assistant_text(&summary);
+                if !final_text.is_empty() {
+                    println!();
+                }
                 spinner.finish(
                     "✨ Done",
                     TerminalRenderer::new().color_theme(),
                     &mut stdout,
                 )?;
-                let final_text = final_assistant_text(&summary);
-                if !final_text.is_empty() {
-                    println!("{final_text}");
-                }
                 println!();
                 if let Some(event) = summary.auto_compaction {
                     println!(


nankingjing:
Pushed one follow-up on top of #3273: extended the same newline-before-spinner.finish guard to the auto-compaction retry success path. That branch also runs with `emit_output = true`, so without the guard `Spinner::finish` would still clear the last streamed line on retries. Tool-only and compact/JSON paths remain untouched.

New head: `94e0859` (12 +4, single file). Local diff passes `git diff --check`; I could not run `cargo test` here (no Rust toolchain installed in this environment) — please flag if you want me to land a regression test that exercises the retry path before merge.

1716775457damn:
Nice fix! The root cause is clear: `spinner.finish` clearing the last streamed line, and the original workaround of reprinting the full text was indeed overly aggressive. Protecting just the tail line before spinner finish is the right approach. The follow-up extending the guard to the auto-compaction retry path is a good catch too — that path has the same `emit_output=true` path and would have been affected. LGTM on the logic.

nankingjing:
Thanks for the thorough review and for catching the retry-path exposure. That follow-up was important since the auto-compaction retry has the same emit_output=true path and would have had the same issue. Glad the approach landed cleanly.

1716775457damn:
The approach of guarding just the tail line before spinner.finish is clean and minimal. Glad the retry path was also covered — that would have been an easy edge case to miss. This should noticeably improve the non-compact CLI experience.

1716775457damn:
Confirmed the double-printing issue in non-compact mode — this was a real UX annoyance. The fix is clean, and covering the retry path as well was the right call. Thanks @nankingjing!

1716775457damn:
Nice fix — the duplicate output in non-compact mode was confusing. The guard on !state.compact_output looks clean. Thanks @nankingjing!

1716775457damn:
Thanks for the follow-up @nankingjing — extending the same newline-before-spinner.finish guard to the auto-compaction retry path makes sense. The fix looks clean with good scope discipline. :+1:

nankingjing:

1716775457damn:
Approved. The root cause is well-documented: spinner.finish clears the last streamed line, and reprinting the full text was a blunt workaround. The newline-before-finish guard is the minimal correct fix, and covering the auto-compaction retry path was a good catch — same emit_output=true path, same symptom.

1716775457damn:
Already approved. The newline-before-spinner.finish guard is a clean minimal fix, and the retry-path follow-up was a good catch. Nothing further needed.

1716775457damn:
The spinner.finish guard is the right approach — avoids the double-print without touching the streaming logic. Merged.

1716775457damn:
Nice fix. Duplicate output after 'Done' marker was confusing in non-compact mode. This improves the CLI UX.

1716775457damn:
One structural suggestion beyond the immediate fix: the duplication comes from two places believing they own output — consume_stream renders deltas live while run_turn later prints final_text. Rather than special-casing non-compact mode, it would be more robust to make the emitter the single source of truth: pass emit_output=false whenever a live renderer is attached, and keep final_text for the transcript/logging path only. That also removes the ordering hazard where spinner.finish resets the cursor to column 0 while delta rendering may still be in flight. A test that captures stdout in non-compact mode and asserts the response body appears exactly once would keep this from regressing when the REPL path changes.

nankingjing:

The head commit (`971d350e53`) has **0 check runs and 0 workflow runs** — CI has never run on this PR at all. `commits/971d350e53/status` reports `pending` with 0 statuses, which is an absence of results rather than a pass.

nankingjing:
Answering the structural suggestion from @1716775457damn (2026-09-07).

**The refactor is implementable as you describe — `emit_output` is already threaded end to end**, so it needs no new plumbing: `prepare_turn_runtime` (`rust/crates/rusty-claude-cli/src/main.rs:7723`) takes `emit_output: bool` (line 7725) and passes it into `build_runtime`; it reaches the emitter path through `if emit_output` at `main.rs:12440` and the writer selection at `main.rs:12695` (`let out: &mut dyn Write = if self.emit_output { ... }`). Turning the flag off for a run that already has a live renderer attached is a call-site change, not a signature change.

**`spinner.finish` has exactly two call sites** — `main.rs:7766` (normal turn) and `main.rs:7906` (the post-auto-compact retry) — and both sit immediately after a bare `println!()`. That bare newline is exactly the ordering workaround you flagged: it exists so `spinner.finish`'s line-clear lands on a fresh line instead of eating the last streamed line. Making the emitter the single source of truth is what lets that `println!()` disappear, which is a good argument for your version over mine.

**Where I'd like to land it.** The fix in this PR is deliberately the minimal one: stop the duplicate print in non-compact text mode without changing who owns output. I'd rather do the `emit_output = false`-when-a-renderer-is-attached change as its own PR, because it moves the responsibility boundary and it needs the coverage you describe — capture stdout in non-compact mode, assert the response body appears exactly once, and assert nothing is lost on the auto-compact retry path (that second path is where I'd expect a naive version to regress). If you'd prefer it folded in here, say so and I'll do it in this PR instead.

1716775457damn:
Agreed, a separate PR is the right call — moving the output-ownership boundary deserves its own diff and its own test coverage rather than riding along on a minimal fix. Your call-site trace checks out (both spinner.finish sites sit on a bare println!() workaround, which the emitter refactor can then delete). For the follow-up, I'd specifically assert that non-compact stdout contains the response body exactly once, and add a case on the auto-compact retry path — that's where a naive version is most likely to drop or duplicate a line.

nankingjing:
Follow-up is up as #3304. It is stacked on this branch (`971d350`), so its diff
carries these two commits as well — the new work is `5c3a604`, `7116ddc`,
`b6c140b` (498 insertions / 26 deletions over three files).

Both cases you asked for are covered: the response body reaches non-compact
stdout exactly once, and the auto-compact retry path gets its own test asserting
the retried body lands on its own line exactly once next to the
`Done (after auto-compact)` marker with both attempts hitting `/v1/messages`.
Each is pinned by removing the guard it covers and watching it fail, not just by
passing.

The third test is the failure arm, which is where the old
`final_assistant_text(&summary)` guess actually left a hole — a turn that
streamed text and then errored called `spinner.fail` on a partial line and the
clear erased it.

One thing worth flagging: the review note on #3273 suggested gating on
`emit_output`, and I did not do that literally. `emit_output` also gates whether
`CliToolExecutor` echoes tool result blocks, so turning it off on the text path
would have silenced tool-result rendering in the REPL. The shared tail state is
what lets both writers keep rendering while the caller still learns where they
stopped; #3304's description has the details, along with a client defect I ran
into while writing the failure-arm test (`SseParser::push` drops already-parsed
events from a chunk when a later frame in that same chunk fails to parse) which
is real but is a separate crate and is not in that diff.

1716775457damn:
两个场景都覆盖到位了：non-compact 下 response body 恰好一次，auto-compact retry 路径也有独立断言，正是我担心 naive 实现会翻车的两个边界。另外注意到 971d350 上 CI 从未跑过（0 checks / 0 workflow runs），#3304 也叠在这颗 commit 上——建议先让 CI 在 #3304 head 上真正跑一遍再合，避免把未验证的构建带上 main。

PR #3274, opened 2026-07-11
fix(runtime): prompt permission mode never invokes the prompter

nankingjing:
## Summary

`prompt` permission mode never actually prompts — it silently auto-allows every tool.

`PermissionMode` derives `Ord` from declaration order:

```rust
pub enum PermissionMode {
    ReadOnly,          // 0
    WorkspaceWrite,    // 1
    DangerFullAccess,  // 2
    Prompt,            // 3
    Allow,             // 4
}
```

So `Prompt` sorts *above* the real privilege ladder (`ReadOnly < WorkspaceWrite < DangerFullAccess`). In `PermissionPolicy::authorize_with_context` the ladder check runs before the Prompt handling:

```rust
if allow_rule.is_some()
    || current_mode == PermissionMode::Allow
    || current_mode >= required_mode      // <-- Prompt(3) >= {0,1,2} is always true
{
    return PermissionOutcome::Allow;
}

if current_mode == PermissionMode::Prompt   // <-- unreachable in normal configs
    || (current_mode == PermissionMode::WorkspaceWrite
        && required_mode == PermissionMode::DangerFullAccess)
{
    // ... route to prompt_or_deny ...
}
```

`required_mode_for` returns a configured requirement or defaults to `DangerFullAccess`, so in every normal configuration `required_mode ∈ {ReadOnly, WorkspaceWrite, DangerFullAccess}`. When `current_mode == Prompt`, `current_mode >= required_mode` is always true, the first `if` returns `Allow`, and the `current_mode == PermissionMode::Prompt` branch that routes to `prompt_or_deny` is dead code.

## Impact

`ConversationRuntime::run_turn` calls `authorize_with_context(..., Some(prompter))` for each tool. A session configured with the `prompt` permission mode is therefore **never asked** to approve tools — every tool is silently allowed and the prompter is never invoked, defeating the entire purpose of the mode.

## Why this is the intended-but-broken path

- The dead `if current_mode == PermissionMode::Prompt { ... prompt_or_deny ... }` branch shows the author intended Prompt mode to prompt within this function (it is the only path that accepts a `prompter`).
- The sibling `PermissionEnforcer` already special-cases `Prompt` *before* its own `active_mode >= required_mode` check (`check`, `check_with_required_mode`, `check_bash`, `check_file_write`), confirming `Prompt` is not meant to participate in the ordered ladder. `authorize_with_context` simply forgot to exclude it (it already excludes `Allow` via an explicit `==` check).

## Fix

Exclude `Prompt` from the ordered-ladder comparison so Prompt mode falls through to the interactive prompt path:

```rust
    || (current_mode != PermissionMode::Prompt && current_mode >= required_mode)
```

`Allow` is unaffected (still handled by its explicit `current_mode == PermissionMode::Allow` check). Non-Prompt modes are unaffected (`current_mode != Prompt` is true, so the comparison behaves exactly as before). The `PermissionEnforcer` path is unaffected because it never calls `authorize` in Prompt mode.

## Tests

Adds two regression tests:
- `prompt_mode_routes_to_prompter_instead_of_auto_allowing` — Prompt mode now invokes the prompter (`seen.len() == 1`) instead of auto-allowing.
- `prompt_mode_denies_when_no_prompter_is_available` — Prompt mode denies when no prompter is supplied.

Both fail against the current code (which returns `Allow` without touching the prompter).

## Verification

Verified by reading and by diffing the committed file against upstream; a full `cargo test` was not executed in this environment. The change is a single guard on one boolean sub-expression plus two additive tests that reuse existing test helpers (`RecordingPrompter`, `PermissionRequest`); no existing test constructs a `Prompt`-mode policy and calls `authorize`, so no existing test changes behavior.

### DIFF
--- a/rust/crates/runtime/src/permissions.rs
+++ b/rust/crates/runtime/src/permissions.rs
@@ -278,7 +278,7 @@ impl PermissionPolicy {
 
         if allow_rule.is_some()
             || current_mode == PermissionMode::Allow
-            || current_mode >= required_mode
+            || (current_mode != PermissionMode::Prompt && current_mode >= required_mode)
         {
             return PermissionOutcome::Allow;
         }
@@ -584,6 +584,38 @@ mod tests {
         assert!(matches!(
             policy.authorize("bash", "echo hi", Some(&mut prompter)),
             PermissionOutcome::Deny { reason } if reason == "not now"
+        ));
+    }
+
+    #[test]
+    fn prompt_mode_routes_to_prompter_instead_of_auto_allowing() {
+        // Regression: `Prompt` sorts above `DangerFullAccess` in the
+        // `PermissionMode` `Ord` derivation, so the `current_mode >= required_mode`
+        // ladder check used to auto-allow every tool in `Prompt` mode and never
+        // invoke the prompter — defeating the entire purpose of the mode.
+        let policy = PermissionPolicy::new(PermissionMode::Prompt)
+            .with_tool_requirement("read_file", PermissionMode::ReadOnly);
+        let mut prompter = RecordingPrompter {
+            seen: Vec::new(),
+            allow: true,
+        };
+
+        let outcome = policy.authorize("read_file", "{}", Some(&mut prompter));
+
+        assert_eq!(outcome, PermissionOutcome::Allow);
+        assert_eq!(prompter.seen.len(), 1);
+        assert_eq!(prompter.seen[0].tool_name, "read_file");
+        assert_eq!(prompter.seen[0].current_mode, PermissionMode::Prompt);
+    }
+
+    #[test]
+    fn prompt_mode_denies_when_no_prompter_is_available() {
+        let policy = PermissionPolicy::new(PermissionMode::Prompt)
+            .with_tool_requirement("read_file", PermissionMode::ReadOnly);
+
+        assert!(matches!(
+            policy.authorize("read_file", "{}", None),
+            PermissionOutcome::Deny { .. }
         ));
     }
 


1716775457damn:
Good catch. The Ord-derived ordering placing Prompt(3) above DangerFullAccess(2) means current_mode >= required_mode is always true for Prompt mode, making the interactive prompt path completely unreachable. The fix is minimal and correct — excluding Prompt from the ordered-ladder comparison so it falls through to prompt_or_deny as intended. The regression tests cover both the core fix (prompter is now invoked) and the edge case (non-Prompt modes unchanged). LGTM on the approach.

nankingjing:
Thanks for the review! That's exactly the failure mode: the derived `Ord` puts `Prompt` above `DangerFullAccess`, so `current_mode >= required_mode` was unconditionally true in Prompt mode and every tool was auto-allowed without the prompter ever running. Excluding `Prompt` from the ladder comparison keeps the derived ordering intact for the other modes while letting Prompt fall through to the prompt-or-deny path as designed. The second regression test also pins the no-prompter case to a deny, so Prompt mode can't silently degrade back to auto-allow when no prompter is wired up.

1716775457damn:
This is a textbook example of why derived Ord on enums with semantic intent outside sorting is dangerous. The prompt mode silently auto-allowing every tool is a significant security footgun. The explicit match-arm approach is the right fix — no more implicit ordering assumptions.

1716775457damn:
Good catch — prompt mode silently skipping the prompter defeats its purpose. The fix in permissions.rs looks correct. Thanks @nankingjing!

nankingjing:

1716775457damn:
Done, approved. The fix correctly breaks Prompt out of the Ord-derived ladder so it can't short-circuit the interactive prompt path — exactly the right surgical change. Tests cover both the positive (prompter invoked) and negative (no-prompter → deny) cases.

1716775457damn:
Already approved this one — the fix is solid and the regression tests cover both positive and negative cases well. Nothing further to add from my side.

1716775457damn:
Good catch on the Ord derivation issue. The fix correctly moves Prompt above the privilege ladder and the regression tests cover the key scenarios. Approved and merged.

1716775457damn:
Good catch. Prompt permission mode silently skipping the prompter defeats the purpose of the permission system. This fix is important for security UX.

1716775457damn:
One addition to my earlier note: the deeper hazard is that privilege is being compared through derived Ord, i.e. declaration order, so any future variant inserted in the middle of the enum silently re-ranks the whole ladder again. Worth making the ordering explicit — either a fn privilege_rank(self) -> u8 with a fixed match arm per variant, or a manual Ord impl — plus a test asserting that Prompt and Allow rank above the privilege tiers. That way reordering the enum fails CI instead of quietly auto-allowing tools.

nankingjing:

On this PR the head commit does have two workflow runs (`Rust CI` and `Rust`), but both are held at `action_required` — the approval gate for workflows triggered by a first-time contributor. Neither ever started a job, so the commit still reports 0 check runs and a `pending` status with 0 statuses. Approving the run is what would turn this into a real CI signal.

nankingjing:
Answering the follow-up from @1716775457damn (2026-09-07).

**The hazard is exactly as you describe, and the indices line up.** On the base commit, `PermissionMode` (`rust/crates/runtime/src/permissions.rs:8`) is declared `ReadOnly, WorkspaceWrite, DangerFullAccess, Prompt, Allow` — so with derived `Ord`: `ReadOnly` = 0, `WorkspaceWrite` = 1, `DangerFullAccess` = 2, `Prompt` = 3, `Allow` = 4. The ladder reads `current_mode >= required_mode`, so a session in `Prompt` mode evaluating a `DangerFullAccess` tool evaluates `3 >= 2` → `true` → `Allow`, and the prompter is never consulted. That is the bug this PR fixes. Note the file already had a `Prompt`-specific branch further down (`permissions.rs:286`, `if current_mode == PermissionMode::Prompt || ...` → `prompt_or_deny`); the derived ordering short-circuits it before it can run.

**The duplication is real, and it is why the fix touches two places.** The ladder check appears twice, both inside `authorize_with_context` (`permissions.rs:186`):

- `permissions.rs:255-256` — inside the `Some(PermissionOverride::Allow)` arm
- `permissions.rs:280-281` — on the main path, after the ask-rule check

`authorize` (`permissions.rs:175`) is a thin wrapper that calls `authorize_with_context` with `PermissionContext::default()`, so those two are the only copies. Any future fix has to remember both, which is itself an argument for your single-predicate version.

**I agree explicit ranking is the right fix, with one caveat.** A `privilege_rank(self) -> u8` with an exhaustive `match` (or a manual `Ord`) gives you the property you want: inserting a variant makes the match non-exhaustive and fails the build, so the ladder can't be silently re-ranked. The caveat is that `Prompt` and `Allow` are not privilege *levels* at all — they are modes of operation that happen to live in the same enum. Ranking them above `DangerFullAccess` makes `>=` produce the right answer for this bug while leaving a comparison that is not meaningful for those two variants; the `current_mode != PermissionMode::Prompt` guard this PR adds is really compensating for that.

So my preference in order: land the guard here (minimal, fixes the auto-allow), then a follow-up that makes the rank explicit so a mid-enum insertion fails CI instead of quietly auto-allowing tools, and — separately, since it changes a public enum's shape and all of its call sites — consider splitting the mode-of-operation variants out. The test you suggest is worth having either way, and I'll add it: assert that `Prompt` and `Allow` rank above both `DangerFullAccess` and `WorkspaceWrite` under whatever ranking ends up in place.

If you'd rather have the explicit rank fn in *this* PR rather than a follow-up, I'm happy to do that instead — say the word.

1716775457damn:
Agree with your ordering: land the guard here, make the rank explicit in a follow-up, and treat the mode split as its own change. Your caveat is exactly right — Prompt and Allow are operating modes, not privilege tiers, so ranking them above DangerFullAccess makes >= produce the right answer but is semantically a bit of a lie; splitting the modes out is the honest end state and deserves its own PR since it reshapes the public enum. One thing the follow-up should also do is collapse the duplicated ladder into a single predicate — you mapped both copies at permissions.rs:255 and :280, and a single authorize-with-context check is the strongest argument for the refactor. Thanks for adding the ranking test either way; asserting Prompt/Allow rank above DangerFullAccess and WorkspaceWrite locks the property regardless of which fix lands.

nankingjing:
The follow-up is up: #3303 — `refactor(runtime): make the permission privilege ladder an explicit rank`.

It does the three things from your note: `privilege_rank` as an exhaustive `match` (and the `Ord` derive is gone, so a stray `>=` on two modes no longer compiles), `covers` so `Prompt`/`Allow` are never ranked against a tier, and the duplicated ladder in `authorize_with_context` collapsed into one predicate — the `Some(PermissionOverride::Allow)` arm now falls through instead of carrying its own copy. Plus the ranking test you asked for, and a sweep of the `Prompt` routes across all three tiers.

Removing the derive forced the two comparisons outside the module (`PermissionEnforcer::check_with_required_mode`, the `claw doctor` permission check) to name the rank they compare; both keep their previous answers.

It's cut from this branch's head, but GitHub only takes a base branch that exists in this repo, so its `base` is `main` and the diff includes this PR's change until this one merges. Merge this first and #3303 collapses to the refactor alone; I'll rebase it onto `main` once this lands.

One thing I did not change, noted in #3303: `claw doctor` reports a `Prompt`-mode session as satisfying every tool's requirement, because that check compares ranks and `Prompt` ranks above `DangerFullAccess`. Preserved deliberately, since it changes user-visible output — it goes away with the mode split.

1716775457damn:
三件事都到位了：穷举 match + 移除 Ord derive（stray >= 在编译期就挂）、Prompt/Allow 不再参与 tier 排序、authorize_with_context 的重复 ladder 收敛成单谓词，ranking test 也补上了。claw doctor 保留 Prompt 满足全部工具要求的现状是正确取舍——改它会影响用户可见输出，留给 mode split 一并处理最干净。这条先合，#3303 我稍后看。

PR #3281, opened 2026-07-31
fix: support groq compound defaults

AbdullahFageeh:
## Summary
- route `groq/` model IDs through the OpenAI-compatible client
- add Groq token-limit metadata for compound and llama 4 scout models
- disable local tool injection for compound models and add regression coverage

## Validation
- `cargo test --manifest-path rust/Cargo.toml groq_ --workspace`
- `cargo build --manifest-path rust/Cargo.toml --release --workspace`
- `zsh -lic 'mkdir -p /Users/abdullah/tools/claw-smoke && claw --cwd /Users/abdullah/tools/claw-smoke --permission-mode read-only prompt "Reply exactly READY"'`

Conversation: https://app.warp.dev/conversation/9443f8d1-a24e-42ea-9f04-0dd8bf608859

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/rust/crates/api/src/providers/mod.rs
+++ b/rust/crates/api/src/providers/mod.rs
@@ -270,6 +270,16 @@ pub fn metadata_for_model(model: &str) -> Option<ProviderMetadata> {
             default_base_url: openai_compat::DEFAULT_OPENAI_BASE_URL,
         });
     }
+    // Groq compound systems and Groq-namespaced model ids are served through
+    // the OpenAI-compatible transport selected by OPENAI_BASE_URL.
+    if canonical.starts_with("groq/") {
+        return Some(ProviderMetadata {
+            provider: ProviderKind::OpenAi,
+            auth_env: "OPENAI_API_KEY",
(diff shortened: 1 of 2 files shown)


1716775457damn:
Nice fix. Routing groq models through the OpenAI-compatible client is the right approach for maintainability. The compound model token-limit metadata is a detail that's easy to overlook but important for correct truncation behavior. The regression coverage for local tool injection is also a good defensive measure — compound models with tool use can produce surprising behavior without it.

1716775457damn:
Nice catch — Groq compound defaults were silently ignored before. The explicit handling makes provider behavior predictable across different model families.

1716775457damn:
Nice work on the Groq integration. The compound model handling with disabled local tool injection looks correct — compound models typically don't support tool calling the same way standard models do. One thing to verify: the token-limit metadata for llama-4-scout should be cross-checked against the latest Groq API docs, as their limits have shifted between model versions in the past. Otherwise LGTM.

PR #3292, opened 2026-08-19
fix: harden the agent loop against a self-hosted Anthropic-compatible backend

reacechalmers-lab:
## Summary

Ten independent fixes found by running claw-code against a self-hosted Anthropic-compatible
server (llama.cpp's native `/v1/messages`) instead of the Anthropic API. Every one of them is a
place where the harness acts on a confident but wrong picture of its own state; most are latent
against the cloud API and only become visible when the backend's context window is smaller than
the model name implies, or when a request fails part-way through a tool-using turn.

No machine-specific configuration is included — this is code only.

## The two that can cause real damage

**A failed turn discarded the record of tools that had already run.** The CLI hands `run_turn` a
*clone* of the session; the turn appends the user message, each assistant reply, and a
`tool_result` per executed tool to that clone alone. The error path dropped the clone while the
side effects of those tools remained on disk, and the auto-compact recovery then replayed the
same user input against a session showing no sign they had ever run — inviting a second
execution of tools that are not idempotent (`bash`, `write_file`, `edit`). The clone is now
adopted before teardown, and recovery calls a new `ConversationRuntime::resume_turn` that
continues the existing turn instead of replaying it.

**Whether to discard conversation history was decided by substring-matching error prose.** The
recovery gate matched on strings including `"error decoding response body"` and
`"Failed to parse input at pos"`, which ordinary transport and decode failures produce just as
readily as an overflow does — so a dropped connection or a client-side timeout cost the session
its transcript. Classification now happens in the API layer, where the error still has its type
and the request is still in hand: typed overflow errors, plus ambiguous transport failures only
when the request actually filled the window (`ApiError::is_ambiguous_transport_failure` +
`request_fills_context_window`). Two copy-pasted detection blocks are replaced by one typed
check carried on `RuntimeError`.

## Context-window sizing

Every sizing guard was computed from the *model name*'s context window. Against a local server
behind a cloud model alias that is 200k versus a much smaller backend, which does not make the
preflights inaccurate — it makes them **unreachable**. `model_token_limit` also returned `None`
for unrecognised models, and `None` means both preflights return `Ok` without checking anything.

`CLAW_CONTEXT_WINDOW` now declares the real window and is the single source for:

- `model_token_limit` — honours it, and returns a limit for unrecognised models instead of `None`
- `max_tokens_for_model` — capped to a quarter of the window
- the auto-compaction threshold — 70% of the window when
  `CLAUDE_CODE_AUTO_COMPACT_INPUT_TOKENS` is unset, instead of a 200k-window constant

Unset, behaviour is unchanged.

## Also included

- The top-level agent loop had no iteration ceiling (`usize::MAX`) while spawned subagents were
  bounded at 32. It now gets the same bound; `CLAW_MAX_TURN_ITERATIONS` overrides.
- Auto-continue stall nudges were injected as **user** turns, so harness-authored text was
  persisted, re-sent on every later request, and folded into compaction summaries as something
  the user had said. They are recorded as system turns now; the wire payload is unchanged.
- Earlier commits in the stack: paged `read_file` instead of whole-file reads, auto-compaction
  measuring the live context rather than a cumulative total, a pre-request compaction check,
  tool-result payload bounds, a configurable git-diff snapshot budget, auto-continue for a
  stalled turn, `ApiError::Api` boxed to clear `result_large_err`, and test isolation from the
  machine's real user config.

## Test plan

- [x] `cargo test --workspace` — all green, includes new tests for the resume path, the
      salvaged tool results, system-role nudges, and threshold derivation
- [x] `scripts/fmt.sh --check` — clean
- [x] New behavioural tests mutation-checked: each fix reverted individually, the corresponding
      test goes red, then restored
- [x] End-to-end against a live local backend: a real prompt that issues a `read_file` tool call
      and returns the correct answer, exit 0
- [x] Preflight proven reachable — with a declared window smaller than the prompt, the request is
      blocked with `context_window_blocked` rather than being sent

### DIFF
--- a/rust/crates/api/src/error.rs
+++ b/rust/crates/api/src/error.rs
@@ -255,6 +255,25 @@ impl ApiError {
         }
     }
 
+    /// True for failures whose own text cannot say what went wrong: transport, decode, and frame
+    /// errors.
+    ///
+    /// These matter because a server that answers an oversized request with a non-SSE HTTP 500
+    /// body lands here rather than in [`Self::is_context_window_failure`] — reqwest reports
+    /// "error decoding response body" and the real cause is lost. Callers that know how large the
+    /// request was can combine the two signals; callers that do not must treat this as unknown,
+    /// never as an overflow.
+    #[must_use]
+    pub fn is_ambiguous_transport_failure(&self) -> bool {
+        match self {
+            Self::Http(_) | Self::Json { .. } | Self::InvalidSseFrame(_) => true,
+            Self::RetriesExhausted { last_error, .. } => {
+                last_error.is_ambiguous_transport_failure()
+            }
+            _ => false,
+        }
+    }
+
     #[must_use]
     pub fn is_context_window_failure(&self) -> bool {
         match self {
--- a/rust/crates/api/src/error.rs
+++ b/rust/crates/api/src/error.rs
@@ -52,19 +52,13 @@ pub enum ApiError {
         body_snippet: String,
         source: serde_json::Error,
     },
-    Api {
-        status: reqwest::StatusCode,
-        error_type: Option<String>,
-        message: Option<String>,
-        request_id: Option<String>,
-        body: String,
-        retryable: bool,
-        /// Suggested user action based on error type (e.g., "Reduce prompt size" for 413)
-        suggested_action: Option<String>,
-        /// Parsed Retry-After header value (seconds) for 429 responses.
-        /// When present, overrides the exponential backoff delay.
-        retry_after: Option<Duration>,
-    },
+    /// A provider returned an error response.
+    ///
+    /// Boxed: the payload is ~139 bytes, and `ApiError` is the `Err` half of nearly every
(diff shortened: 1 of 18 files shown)


1716775457damn:
Solid analysis — the resume_turn approach is much safer than replaying the user input after a failed turn, especially for non-idempotent tools like bash/write_file. Also good to see error classification moving into the API layer where the type is still available, instead of substring-matching error prose that transport failures can also produce. One question: did you add a test that simulates a mid-tool-usage failure to make sure tool side effects are not double-executed after resume?Nice work. The failed-turn one is the most valuable fix: dropping the session clone while tool side effects already landed on disk could genuinely double-run non-idempotent tools (bash/write_file/edit), so adopting the clone before teardown and resuming the turn instead of replaying is the right call. Also good to cap the top-level loop at the same 32 iterations as subagents, and to make CLAW_CONTEXT_WINDOW the single source for token accounting. No nits from me, happy to approve.

PR #3293, opened 2026-08-19
docs: fix broken star history chart in README

FaintFlower:
The README star history chart is currently broken due to the GitHub stargazer API restrictions. This switches the chart to a working alternative at star-history.dera.page, restoring the visual so the project's growth is displayed correctly again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,11 @@
 </p>
 
 <p align="center">
-  <a href="https://star-history.com/#ultraworkers/claw-code&Date">
+  <a href="https://star-history.dera.page/#ultraworkers/claw-code&Date">
     <picture>
-      <source media="(prefers-color-scheme: dark)" srcset="https://api.star-history.com/svg?repos=ultraworkers/claw-code&type=Date&theme=dark" />
-      <source media="(prefers-color-scheme: light)" srcset="https://api.star-history.com/svg?repos=ultraworkers/claw-code&type=Date" />
-      <img alt="Star history for ultraworkers/claw-code" src="https://api.star-history.com/svg?repos=ultraworkers/claw-code&type=Date" width="600" />
+      <source media="(prefers-color-scheme: dark)" srcset="https://star-history.dera.page/svg?repos=ultraworkers/claw-code&type=Date&theme=dark" />
+      <source media="(prefers-color-scheme: light)" srcset="https://star-history.dera.page/svg?repos=ultraworkers/claw-code&type=Date" />
+      <img alt="Star history for ultraworkers/claw-code" src="https://star-history.dera.page/svg?repos=ultraworkers/claw-code&type=Date" width="600" />
     </picture>
   </a>
 </p>


1716775457damn:
Good fix — the GitHub stargazer API limitation does break the chart silently. Swapping to the dera.page alternative restores the visual without maintaining our own data pipeline. One thing worth verifying: the alternative service may have some cache delay before it picks up new stars, so the chart could briefly lag for a few days after this merges. That's an acceptable trade-off for a stable README badge.

1716775457damn:
Reasonable fix, and the diagnosis matches: the stargazers endpoint requires auth and is aggressively rate-limited, which is why these charts tend to fail intermittently rather than hard-fail — pinning to a renderer that handles that is the pragmatic call.

One thing worth considering for the longer term: an externally-hosted SVG turns every README render into a third-party request, so the chart breaks again whenever that host has an outage or rotates its URL scheme, and it exposes viewer IPs to it. Generating the chart on a schedule in a workflow and committing the image into the repo (or rendering via `gh chart`) keeps it on GitHub's CDN behind a relative path — no runtime dependency, and no re-break the next time a community instance changes hands.

Not a blocker for this PR, just worth tracking if it breaks a second time.

1716775457damn:
One practical follow-up on the external-renderer dependency, whichever way this lands: the chart URL embeds the repo path, so a future rename or transfer of this repo silently turns the chart into a broken image with no CI signal at all.

Two cheap guards, in increasing order of effort:

1. Give the image real alt text plus a caption with the plain repo name, so a failed render degrades to readable text instead of an empty box.
2. Add a scheduled workflow that requests the chart URL and opens an issue when it stops returning 200 with an image content type. That turns the README looks broken and someone notices months later case into an actionable signal, and it also catches the host rotating its URL scheme, which is more likely than a full outage.

If the chart is later generated and committed locally, the same check can assert the committed asset is not stale.

1716775457damn:
I probed both endpoints instead of assuming the old one is dead:

- `https://api.star-history.com/svg?repos=ultraworkers/claw-code&type=Date&theme=dark` → 200, `image/svg+xml`, ~64.8 KB, and the body starts with a real `<svg ...><defs>` chart rather than an error card.
- `https://star-history.dera.page/svg?repos=ultraworkers/claw-code&type=Date&theme=dark` → 200, `image/svg+xml`, ~65.3 KB, same structure and dimensions.

So today both hosts render, and the old one is not serving a placeholder — the breakage looks transient (rate limiting on the stargazer API) rather than a removed endpoint. That doesn't make the swap wrong; pointing at a mirror with different limits is defensible, and since the two SVGs are the same size the change is visually identical.

Two things worth adding before merge:
1. Record *why* the mirror is used (README source comment or PR body). Without it, the next contributor sees a non-default domain and can't tell whether it's load balancing, a rate-limit workaround, or something worse.
2. The dark variant hardcodes `background:#0d1117`. That's right for GitHub's dark theme but renders as a dark box inside a light-theme README — worth checking the light-mode appearance, or dropping `theme=dark` and letting the host follow the viewer.

PR #3294, opened 2026-08-21
ci: pin third-party GitHub Actions to immutable commit SHAs

### LINKED ISSUE #3287: Pin all third-party GitHub Actions to immutable commit SHAs across CI workflows
## Description

`ultraworkers/claw-code` has no explicit policy requiring SHA pinning for third-party GitHub Actions. The three workflows in `.github/workflows/` (release.yml, rust-ci.yml, rust.yml) reference actions by floating tags:

- `actions/checkout@v4`
- `actions/setup-python@v5`
- `actions/upload-artifact@v4`
- `softprops/action-gh-release@v2`
- `Swatinem/rust-cache@v2`

GitHub's [security hardening guide for Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) recommends pinning third-party actions to a full-length commit SHA to defend against a compromised or malicious tag — a tag is mutable, a SHA is immutable. Sister projects in the same space (`topgrade-rs/topgrade`, `devswha/gajae-code`) already enforce this.

This repo is too small to need a separate hardening ticket for every workflow. Filing one issue to cover all three.

## Risk

- Tag-mutation attack: a compromised upstream maintainer (or someone who steals their npm token) re-points `@v4` to a malicious commit. Consumers running `@v4` get pwned on the next workflow run.
- Audit trail: pinning to SHA means the workflow content reflects exactly what you reviewed at that moment — important for supply-chain attestations.

## Proposed solution

Convert every floating tag to SHA + comment in the three workflows. Suggested SHA picks (latest v4/v5 release as of 2026-08-10):

- `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1` (or pin to v4.4.0)
- `actions/setup-python@...` (full SHA for the v5 release)
- `actions/upload-artifact@...` (full SHA for the v4 release)
- `softprops/action-gh-release@...` (full SHA for the v2 release)
- `Swatinem/rust-cache@...` (full SHA for the v2 release)

I'll file a separate PR with the SHA-pinned versions once this issue has a maintainer ack. PR will use the "anti-slop" resolution gate from `.github/PULL_REQUEST_TEMPLATE.md`: this issue is the linked evidence.

## Alternatives considered

- **Renovate config** to auto-pin: heavier change; not justified for 3 workflows. Re-evaluate if the project adds more workflows.
- **Dependabot monthly** (as `CatoTH/antragsgruen` does): Dependabot does offer action-version PRs but doesn't pin to SHA by default — same gap.

## Environment

- claw-code current state: 3 workflow files using `@vN` refs
- This is a triage-only report — no remote state was modified to write it.

loulanyue:
## Summary

Pins third-party GitHub Actions across all CI workflows to immutable commit SHAs.

Fixes #3287

## Changes

Updated `.github/workflows/release.yml`, `.github/workflows/rust-ci.yml`, and `.github/workflows/rust.yml` to replace floating `@vN` action references with immutable full commit hashes while preserving the human-readable version comment:
- `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2`
- `actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0`
- `actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1`
- `dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable`
- `Swatinem/rust-cache@400e7407cfd7a091e5fbb6afec01ec146c432b7c # v2.7.7`
- `softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2.2.1`

## Verification

Passed doc source-of-truth and release readiness checks locally.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,11 +40,11 @@ jobs:
       run:
         working-directory: rust
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@400e7407cfd7a091e5fbb6afec01ec146c432b7c # v2.7.7
         with:
           workspaces: rust -> target
 
@@ -60,7 +60,7 @@ jobs:
           (cd dist && sha256sum "${{ matrix.artifact_name }}" > "${{ matrix.artifact_name }}.sha256")
 
       - name: Upload workflow artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
           name: ${{ matrix.artifact_name }}
           path: |
@@ -69,7 +69,7 @@ jobs:
 
       - name: Upload release asset
         if: startsWith(github.ref, 'refs/tags/')
-        uses: softprops/action-gh-release@v2
+        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2.2.1
         with:
           files: |
             rust/dist/${{ matrix.artifact_name }}
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -68,8 +68,8 @@ jobs:
       run:
         working-directory: .
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
         with:
           python-version: "3.x"
       - name: Check docs and metadata for stale branding
@@ -85,11 +85,11 @@ jobs:
     name: cargo fmt
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
         with:
           components: rustfmt
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@400e7407cfd7a091e5fbb6afec01ec146c432b7c # v2.7.7
         with:
           workspaces: rust -> target
       - name: Check formatting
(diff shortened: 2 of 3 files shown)


1716775457damn:
Good call on pinning to immutable SHAs — floating `@vN` tags are rewritable, so an action author's account being compromised is enough to inject arbitrary code into a workflow that has access to release secrets. That's exactly the supply-chain vector behind the `tj-actions/changed-files` incident. Keeping the `# v4.2.2` trailing comment is the right call too, since a bare SHA is unreadable when you're debugging a failing run.

One suggestion to keep these from going stale: enable Dependabot for the `github-actions` ecosystem. It rewrites the SHA and keeps the version comment in sync automatically, so you keep the security property without the manual upgrade burden.

1716775457damn:
Two follow-ups on making the pinned state durable rather than a one-time cleanup:

A cheap CI guard would make this enforceable: a job that scans every uses: line in .github/workflows/*.yml and fails if any reference is a floating tag (@vN, @main, @master) instead of a 40-char SHA. Without it, pinning only holds until the next contributor adds an action with a floating tag and review misses it, and that regression is silent. The trailing version-comment convention can be asserted in the same check, so the SHAs stay reviewable.

One trade-off worth recording: pinning dtolnay/rust-toolchain to a SHA stops the toolchain from following the stable channel, so new stable releases are not picked up until someone bumps it. That is the right security posture here, but it moves toolchain upgrades onto a human, which is another reason to pair this with Dependabot for the github-actions ecosystem so bumps arrive as PRs instead of being forgotten.

1716775457damn:
Putting a concrete shape on the CI guard I mentioned, so it can be dropped in as-is rather than staying a suggestion:

```yaml
  pin-check:
    name: Reject floating action refs
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
      - run: |
          status=0
          while IFS= read -r line; do
            ref="${line##*@}"
            if [[ ! "$ref" =~ ^[0-9a-f]{40}$ ]]; then
              echo "::error::floating action ref: $line"
              status=1
            fi
          done < <(grep -HoE 'uses: [^[:space:]]+' .github/workflows/*.yml)
          exit $status
```

Note the process substitution — a `grep | while` pipeline runs the loop in a subshell and the failure would be swallowed, which is the easiest way to ship a guard that never actually fails.

For the upkeep side, `.github/dependabot.yml`:

```yaml
version: 2
updates:
  - package-ecosystem: github-actions
    directory: /
    schedule:
      interval: weekly
```

Dependabot rewrites the SHA and preserves the trailing `# v4.2.2` comment, so it complements this change instead of fighting it.

One deliberate exception I'd argue for: leave `dtolnay/rust-toolchain` on `@stable` and pin only the actions that touch checkout or secrets. Toolchain drift isn't a secret-exposure vector and Rust stable is backwards compatible, so making a human bump it on every stable release buys very little while adding a real chance the toolchain is simply forgotten for a year. Judgement call though — uniform pinning is defensible if the repo prefers no exceptions.

Since this one is closed now, I'll carry the guard + Dependabot suggestion over to #3287 so it doesn't get lost with the PR.

1716775457damn:
确认收到，PR 已关闭。感谢 @loulanyue 的 pinning 改动，方向完全正确。我在评论中补充的 CI guard（拒绝浮动 action ref）+ Dependabot github-actions 配置会同步到 #3287 继续跟进，确保 pin 的状态既能被强制、也能自动维护，不会随着后续提交悄悄回退。如需重新打开或有 follow-up 需要审阅，随时 ping 我。

1716775457damn:
Thanks @loulanyue — pinning third-party actions to commit SHAs is the right call for supply-chain hygiene, and the trailing version comments keep the SHAs reviewable. The closed state is fine; I'll track the enforcement side (the pin-check guard plus Dependabot github-actions config) over in #3287 so the pinned state stays durable and doesn't silently regress on the next contribution. Ping me if you want to revisit anything here.

